PR #1, opened 2019-11-07
XDC and FASM Yosys plugins for INTERNAL_VREF support.

tmichalak:
The initial versions of the plugins are capable of reading an XDC file with the INTERNAL_VREF settings and generating a FASM file with the features that we currently have in our database that are related to the VREF settings per bank.
This plugin operates on the existing design and modifies its structure based on the content of the XDC (Xilinx Design Constraints) file.
Since the XDC file consists of Tcl commands it is read using Yosys's tcl command and processed by the new XDC commands imported to the Tcl interpreter.

To use the plugins which are compiled as shared libraries they need to be copied to `share/yosys/plugins/` directory and later on in the yosys script they should be loaded with the `plugin command`

The information about available IO banks for a specific part and what tiles they map to will be taken from the part's json.
The file is passed to read_xdc and write_fasm commands via `-json_part` switch and should contain an `iobanks` entry, e.g.:
<pre>
    "iobanks": {
        "14" : "X1Y26",
        "15" : "X1Y78",
        "16" : "X1Y130",
        "34" : "X113Y26",
        "35" : "X113Y78"
    }
</pre>

This information is used during writing out the fasm file as well as for error checking during read_xdc.

Example of Yosys script:
<pre>
plugin -i fasm
plugin -i xdc
read_verilog counter_arty.v
read_xdc -part_json xc7a35tcsg324-1.json arty.xdc
write_fasm -part_json xc7a35tcsg324-1.json counter_arty.fasm
</pre>

In order to merge it to symbiflow-arch-defs I will have to create a conda package and copy the plugins to yosys's environment. 
The next step will be modifying the synthesis script we use in arch-defs
Finally, add a fasm concatenation step that will combine the output of the fasm plugin and the outcome of genfasm to produce the complete FASM.

### DIFF
--- a/bank_tiles.h
+++ b/bank_tiles.h
@@ -19,40 +19,33 @@
  */
 #include "kernel/log.h"
 #include "libs/json11/json11.hpp"
-//#include <sstream>
 
-#define PART_JSON "PRJXRAY_PART_JSON"
 
 USING_YOSYS_NAMESPACE
 // Coordinates of HCLK_IOI tiles associated with a specified bank
 using BankTilesMap = std::unordered_map<int, std::string>;
+BankTilesMap bank_tiles;
 using json11::Json;
 
 // Find the part's JSON file with information including the IO Banks
 // and extract the bank tiles.
-BankTilesMap get_bank_tiles() {
+BankTilesMap get_bank_tiles(const std::string json_file_name) {
 	BankTilesMap bank_tiles;
-	std::string part_json;
-	try {
-		part_json = std::string(getenv(PART_JSON));
-	} catch (...) {
-		log("write_fasm: %s not defined\n", PART_JSON);
-		return BankTilesMap();
+	std::ifstream json_file(json_file_name);
+	if (!json_file.good()) {
+		log_cmd_error("Can't open JSON file %s", json_file_name.c_str());
 	}
-	std::ifstream json_file(part_json);
 	std::string json_str((std::istreambuf_iterator<char>(json_file)),
-                 std::istreambuf_iterator<char>());
+			std::istreambuf_iterator<char>());
 	std::string error;
 	Json json = Json::parse(json_str, error);
 	if (!error.empty()) {
-		log("get_bank_tiles: json parsing failed \n");
-		return BankTilesMap();
+		log_cmd_error("%s\n", error.c_str());
 	}
 	auto json_objects = json.object_items();
 	auto iobanks = json_objects.find("iobanks");
 	if (iobanks == json_objects.end()) {
-		log("get_bank_tiles: IO Banks information missing in the part's json: %s\n", part_json.c_str());
-		return BankTilesMap();
+		log_cmd_error("IO Bank information missing in the part's json: %s\n", json_file_name.c_str());
 	}
 
 	for (auto iobank : iobanks->second.object_items()) {
--- a/bank_tiles.h
+++ b/bank_tiles.h
@@ -25,7 +25,6 @@ USING_YOSYS_NAMESPACE
 // Coordinates of HCLK_IOI tiles associated with a specified bank
 using BankTilesMap = std::unordered_map<int, std::string>;
 BankTilesMap bank_tiles;
-using json11::Json;
 
 // Find the part's JSON file with information including the IO Banks
 // and extract the bank tiles.
@@ -38,7 +37,7 @@ BankTilesMap get_bank_tiles(const std::string json_file_name) {
 	std::string json_str((std::istreambuf_iterator<char>(json_file)),
 			std::istreambuf_iterator<char>());
 	std::string error;
-	Json json = Json::parse(json_str, error);
+	auto json = json11::Json::parse(json_str, error);
 	if (!error.empty()) {
 		log_cmd_error("%s\n", error.c_str());
 	}
--- a/bank_tiles.h
+++ b/bank_tiles.h
@@ -24,7 +24,6 @@
 USING_YOSYS_NAMESPACE
 // Coordinates of HCLK_IOI tiles associated with a specified bank
 using BankTilesMap = std::unordered_map<int, std::string>;
-BankTilesMap bank_tiles;
 
 // Find the part's JSON file with information including the IO Banks
 // and extract the bank tiles.
@@ -53,3 +52,4 @@ BankTilesMap get_bank_tiles(const std::string json_file_name) {
 
 	return bank_tiles;
 }
+
--- a/bank_tiles.h
+++ b/bank_tiles.h
@@ -1,0 +1,63 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
+ *  Copyright (C) 2019  The Symbiflow Authors
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+#include "kernel/log.h"
+#include "libs/json11/json11.hpp"
+//#include <sstream>
+
+#define PART_JSON "PRJXRAY_PART_JSON"
+
+USING_YOSYS_NAMESPACE
+// Coordinates of HCLK_IOI tiles associated with a specified bank
+using BankTilesMap = std::unordered_map<int, std::string>;
+using json11::Json;
+
+// Find the part's JSON file with information including the IO Banks
+// and extract the bank tiles.
+BankTilesMap get_bank_tiles() {
+	BankTilesMap bank_tiles;
+	std::string part_json;
+	try {
+		part_json = std::string(getenv(PART_JSON));
+	} catch (...) {
+		log("write_fasm: %s not defined\n", PART_JSON);
+		return BankTilesMap();
+	}
+	std::ifstream json_file(part_json);
+	std::string json_str((std::istreambuf_iterator<char>(json_file)),
+                 std::istreambuf_iterator<char>());
+	std::string error;
+	Json json = Json::parse(json_str, error);
+	if (!error.empty()) {
+		log("get_bank_tiles: json parsing failed \n");
+		return BankTilesMap();
+	}
+	auto json_objects = json.object_items();
+	auto iobanks = json_objects.find("iobanks");
+	if (iobanks == json_objects.end()) {
+		log("get_bank_tiles: IO Banks information missing in the part's json: %s\n", part_json.c_str());
+		return BankTilesMap();
+	}
+
+	for (auto iobank : iobanks->second.object_items()) {
+		bank_tiles.emplace(std::atoi(iobank.first.c_str()), iobank.second.string_value());
+	}
+
+	return bank_tiles;
+}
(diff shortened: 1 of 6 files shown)


bank_tiles.h
```diff
@@ -0,0 +1,52 @@
...
+
+#define PART_JSON "PRJXRAY_PART_JSON"
+
+USING_YOSYS_NAMESPACE
+// Coordinates of HCLK_IOI tiles associated with a specified bank
+using BankTilesMap = std::unordered_map<int, std::string>;
+
+// Find the part's JSON file with information including the IO Banks
+// and extract the bank tiles.
+BankTilesMap get_bank_tiles() {
```
litghost: I'm not a fan of using environment variables in this way.  Can we have `write_fasm` take an argument that gets passed to `get_bank_tiles`?  Then in the top level TCL we can use the env var, but we leave it outside of the plugin.  It is generally not a good idea to proscribe a configuration strategy in the plugin.

bank_tiles.h
```diff
@@ -0,0 +1,52 @@
...
+	BankTilesMap bank_tiles;
+	std::string part_json;
+	try {
+		part_json = std::string(getenv(PART_JSON));
+	} catch (...) {
+		log("write_fasm: %s not defined\n", PART_JSON);
+		return BankTilesMap();
+	}
+	std::ifstream json_file(part_json);
+	nlohmann::json json;
```
litghost: Doesn't Yosys already have json support?  Why do we need to bring in a new library?

fasm-plugin/fasm.cc
```diff
@@ -0,0 +1,83 @@
...
+		BankTilesMap bank_tiles(get_bank_tiles());
+		if (bank_tiles.size()) {
+			// Generate a fasm feature associated with the INTERNAL_VREF value per bank
+			// e.g. VREF value of 0.675 for bank 34 is associated with tile HCLK_IOI3_X113Y26
+			// hence we need to emit the following fasm feature: HCLK_IOI3_X113Y26.VREF.V_675_MV
+			for (auto cell : top_module->cells()) {
+				if (cell->type != ID(BANK)) continue;
+				int bank_number(cell->getParam(ID(NUMBER)).as_int());
+				int bank_vref(cell->getParam(ID(INTERNAL_VREF)).as_int());
+				*f << "HCLK_IOI3_" << bank_tiles[bank_number] <<".VREF.V_" << bank_vref << "_MV\n";
```
litghost: Still need error checking?

fasm-plugin/fasm.cc
```diff
@@ -0,0 +1,83 @@
...
+			log("write_fasm: No top module detected\n");
+			return;
+		}
+		// Return if no BANK module exists as this means there are no cells
+		if (!design->has(ID(BANK))) {
+			return;
+		}
+
+		BankTilesMap bank_tiles(get_bank_tiles());
+		if (bank_tiles.size()) {
```
litghost: Should generate error or warning if `bank_tiles` is size == 0, but there are VREF directives?

bank_tiles.h
tmichalak: Yeah, we could have that. `get_bank_tiles` is also used in the xdc plugin during get_iobanks so in order to drop the env variable approach I might have to create an xdc frontend, aka `read_xdc`, which will basically behave like the `tcl` command, but will take the json argument. We could also skip the [bank check](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/1/files/c0c9ebee7c4ae4e2ddbad434d1882c9373c7b85c#diff-bf6d652c14576b97319c59ebd0933b30R90), but I would rather keep it.

bank_tiles.h
tmichalak: Yeah, yosys has one indeed, will use it then.

bank_tiles.h
litghost: The other option is to have a function that generates the `get_bank_tiles` data, and pass it to both `read_xdc` and `write_fasm`.  This may have utility if people want to inspect banks anyways.

mithro: I think @litghost 's option sounds good?

fasm-plugin/fasm.cc
```diff
@@ -0,0 +1,83 @@
...
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ *  ---
```
mithro: This comment is no longer accurate?

xdc-plugin/xdc.cc
```diff
@@ -0,0 +1,172 @@
...
+		int internal_vref = 1000 * std::atof(args[0].c_str());
+		if (internal_vref != 600 &&
+				internal_vref != 675 &&
+				internal_vref != 750 &&
+				internal_vref != 900) {
+			log("set_property INTERNAL_VREF: Incorrect INTERNAL_VREF value\n");
+			return;
+		}
+
+		// Create a new BANK module if it hasn't been created so far
```
mithro: Can these be in Verilog file rather than hard coded into this C++ file?

fasm-plugin/fasm.cc
```diff
@@ -0,0 +1,83 @@
...
+		log("\n");
+	}
+
+	void execute(std::ostream *&f, std::string filename,  std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE {
+		size_t argidx = 1;
+		extra_args(f, filename, args, argidx);
+		process_vref(f, design);
+	}
+
+	void process_vref(std::ostream *&f, RTLIL::Design* design) {
```
mithro: I think this function should just search for any module which has a property named something like `FASM_EXTRA` parameter and write that to the fasm file? It shouldn't need to understand banks or vref stuff?

bank_tiles.h
```diff
@@ -0,0 +1,52 @@
...
+#include "kernel/log.h"
+#include "json.hpp"
+
+#define PART_JSON "PRJXRAY_PART_JSON"
+
+USING_YOSYS_NAMESPACE
+// Coordinates of HCLK_IOI tiles associated with a specified bank
+using BankTilesMap = std::unordered_map<int, std::string>;
+
+// Find the part's JSON file with information including the IO Banks
```
mithro: What json file are you using here?

tmichalak: I plan to use the json that we have in prjxray's db for a given part, for example `xc7a35tcsg324-1.json` with an extra entry that we can generate in one of the fuzzers:
<pre>
    "iobanks": {
        "14" : "X1Y26",
        "15" : "X1Y78",
        "16" : "X1Y130",
        "34" : "X113Y26",
        "35" : "X113Y78"
    }
</pre>

bank_tiles.h
tmichalak: The json file can be passed using a command argument.

bank_tiles.h
tmichalak: Switched to json11

fasm-plugin/fasm.cc
tmichalak: Done

fasm-plugin/fasm.cc
tmichalak: Done

fasm-plugin/fasm.cc
tmichalak: Fixed.

fasm-plugin/fasm.cc
```diff
@@ -0,0 +1,89 @@
...
+PRIVATE_NAMESPACE_BEGIN
+
+struct WriteFasm : public Backend {
+	WriteFasm() : Backend("fasm", "Write out FASM features") {}
+
+
+	void help() YS_OVERRIDE {
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    write_fasm filename\n");
```
litghost: Update comment about `-part_json` argument?

fasm-plugin/fasm.cc
tmichalak: Done

fasm-plugin/fasm.cc
tmichalak: Done

xdc-plugin/xdc.cc

bank_tiles.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ */
+#include "kernel/log.h"
+#include "libs/json11/json11.hpp"
+
+
+USING_YOSYS_NAMESPACE
+// Coordinates of HCLK_IOI tiles associated with a specified bank
+using BankTilesMap = std::unordered_map<int, std::string>;
+BankTilesMap bank_tiles;
```
litghost: What is this instance doing here?

bank_tiles.h
```diff
@@ -0,0 +1,56 @@
...
+using json11::Json;
```
litghost: Do not do this, just call the fully qualified name.

xdc-plugin/xdc.cc
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	void execute(std::istream *&f, std::string filename, std::vector<std::string> args, RTLIL::Design*) YS_OVERRIDE {
+                if (args.size() < 2) {
+                        log_cmd_error("Missing script file.\n");
+		}
+                Tcl_Interp *interp = yosys_get_tcl_interp();
+		size_t argidx = 1;
+		if (args[argidx] == "-part_json" && argidx + 1 < args.size()) {
+			bank_tiles = get_bank_tiles(args[++argidx]);
```
litghost: This use of a global variable is not acceptable.  Can you propose other options?

xdc-plugin/xdc.cc
```diff
@@ -0,0 +1,242 @@
...
+		log("\n");
+		log("Inspect IO Bank tiles for the specified part based on the provided JSON file.\n");
+		log("\n");
+	}
+
+	void execute(std::vector<std::string> args, RTLIL::Design* ) YS_OVERRIDE {
+                if (args.size() < 2) {
+                        log_cmd_error("Missing JSON file.\n");
+		}
+		// Check if the part has the specified bank
```
litghost: There is not reason to use the global variable bank_tiles here.

xdc-plugin/xdc.cc
```diff
@@ -0,0 +1,242 @@
...
+#include "libs/json11/json11.hpp"
+#include "../bank_tiles.h"
+
+USING_YOSYS_NAMESPACE
+
+PRIVATE_NAMESPACE_BEGIN
+
+enum SetPropertyOptions { INTERNAL_VREF };
+
+std::unordered_map<std::string, SetPropertyOptions> set_property_options_map  = {
```
litghost: Make this map const?

xdc-plugin/xdc.cc
```diff
@@ -0,0 +1,242 @@
...
+#include "kernel/rtlil.h"
+#include "kernel/log.h"
+#include "libs/json11/json11.hpp"
+#include "../bank_tiles.h"
+
+USING_YOSYS_NAMESPACE
+
+PRIVATE_NAMESPACE_BEGIN
+
+enum SetPropertyOptions { INTERNAL_VREF };
```
litghost: `enum class`?

xdc-plugin/xdc.cc
litghost: Here is an example of what you could do.

Have `ReadXdc` instance `GetPorts`, `GetIOBanks`, etc.  ReadXdc could own the `bank_tiles` instance, and pass a `std::function<const BankTiles&()>  to the constructors of `GetPorts`, `GetIOBanks` etc.  Upon destruction of `ReadXdc`, all passes and Frontend would do unregister from Yosys.

bank_tiles.h
tmichalak: Fixed

xdc-plugin/xdc.cc

xdc-plugin/xdc.cc

xdc-plugin/xdc.cc

xdc-plugin/xdc.cc
tmichalak: Sure, fixed.

xdc-plugin/xdc.cc
```diff
@@ -0,0 +1,268 @@
...
+		log("Read XDC file.\n");
+		log("\n");
+	}
+
+	void execute(std::istream *&f, std::string filename, std::vector<std::string> args, RTLIL::Design*) YS_OVERRIDE {
+                if (args.size() < 2) {
+                        log_cmd_error("Missing script file.\n");
+		}
+                Tcl_Interp *interp = yosys_get_tcl_interp();
+		size_t argidx = 1;
```
litghost: You should clear bank_tiles if -part_json is not supplied to prevent stale data from passing from call to call.

PR #2, opened 2019-12-17
Annotate verilog with XDC parameters.

tmichalak:
This PR annotates the design read from verilog with IOSTANDARD, SLEW, DRIVE and IN_TERM parameters read from XDC.
Currently only single bit ports are supported.
However we need to handle bus ports for Litex.

### DIFF
--- a/xdc-plugin/xdc.cc
+++ b/xdc-plugin/xdc.cc
@@ -87,11 +87,19 @@ struct GetPorts : public Pass {
 		}
 		// TODO handle more than one port
 		port_name = args.at(1);
-		RTLIL::IdString port_id(RTLIL::escape_id(port_name));
+		char port[128];
+		int bit(0);
+		std::string port_signal(port_name);
+		if (sscanf(port_name.c_str(), "%[^[][%d]", port, &bit) == 2) {
+			port_signal = std::string(port);
+		}
+
+		RTLIL::IdString port_id(RTLIL::escape_id(port_signal.c_str()));
 		if (auto wire = top_module->wire(port_id)) {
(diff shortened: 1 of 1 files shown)


mithro:
Generally LGTM.

mithro:
FYI -- @eddiehung

xdc-plugin/xdc.cc
```diff
@@ -186,6 +224,36 @@ struct SetProperty : public Pass {
...
+		}
+		RTLIL::Module* top_module = design->top_module();
+		RTLIL::IdString port_id(RTLIL::escape_id(port_name.c_str()));
+		for (auto cell_obj : top_module->cells_) {
+			RTLIL::Cell* cell = cell_obj.second;
+			RTLIL::IdString cell_id = cell_obj.first;
+			for (auto connection : cell->connections_) {
+				RTLIL::SigSpec cell_signals = connection.second;
+				if (!strcmp(log_signal(cell_signals),port_id.c_str())) {
+					cell->setParam(RTLIL::IdString(RTLIL::escape_id(parameter)), RTLIL::Const(value));
```
mkurc-ant: Should check here if the cell actually has the parameter of that name. For example IBUF cells does not have DRIVE and SLEW parameters

tmichalak: Understood, but I don't know that if there is a generic way of checking the list of available parameters in yosys. I guess we would have to create our own map with primitives and allowed parameters.

tmichalak: Added for OBUF, OBUFDS and OBUFTDS

Issue #3, opened 2019-12-31
XDC plugin missing annotations


mkurc-ant:
The XDC plugins fails to annotate I/O bufs if they are not connected directly to the top-level port. For example, for design like that the annotation works fine:
```
module top (output wire top_o);

OBUF buf (
.I(some_signal),
.O(top_o)
);

...

endmodule
```

But for this one it does not provided that the `intermediate_wire` doesn't get optimized by Yosys:
```
module top (output wire top_o);

wire intermediate_wire;
assign top_o = intermediate_wire;

OBUF buf (
.I(some_signal),
.O(intermediate_wire)
);

...

endmodule
```

The XDC plugin should traverse from the top-level port down (or up) to any cell and annotate it if it is an IO buffer. Right now it considers only direct connections.

tmichalak:
Solved with #4

PR #4, opened 2020-01-09
XDC: Add wires traversing

tmichalak:
This PR solves issue #3 
It has been tested with the following netlist:
<pre>
module top (
        input  clk,
        output [3:0] led,
        inout out_a,
        output [1:0] out_b
);

        wire LD6, LD7, LD8, LD9;
        wire inter_wire, inter_wire_2;
        localparam BITS = 1;
        localparam LOG2DELAY = 25; 

        reg [BITS+LOG2DELAY-1:0] counter = 0;

        always @(posedge clk) begin
                counter <= counter + 1;
        end
        assign led[1] = inter_wire;
        assign inter_wire = inter_wire_2;
        assign {LD9, LD8, LD7, LD6} = counter >> LOG2DELAY;
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_6(.I(LD6), .O(led[0]));
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_7(.I(LD7), .O(inter_wire_2));
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_OUT(.I(LD7), .O(out_a));
        bottom bottom_inst(.I(LD8), .O(led[2]), .OB(out_b));
        bottom_intermediate bottom_intermediate_inst(.I(LD9), .O(led[3]));
endmodule

module bottom_intermediate (
        input I,
        output O
);
        wire bottom_intermediate_wire;
        assign O = bottom_intermediate_wire;
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_8(.I(I), .O(bottom_intermediate_wire));
endmodule

module bottom (
        input I,
        output [1:0] OB, 
        output O
);
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_9(.I(I), .O(O));
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_10(.I(I), .O(OB[0]));
        OBUF #(
                .IOSTANDARD("LVCMOS33"),
                .SLEW("SLOW")
        ) OBUF_11(.I(I), .O(OB[1]));
endmodule
</pre>

for the following XDC:
<pre>
set_property DRIVE I12 [get_ports {led[0]}]
set_property IN_TERM UNTUNED_SPLIT_40 [get_ports {led[1]}]
set_property SLEW FAST [get_ports {led[1]}]
set_property IOSTANDARD SSTL135 [get_ports {led[1]}]
set_property SLEW FAST [get_ports {led[2]}]
set_property IOSTANDARD SSTL135 [get_ports {led[2]}]
set_property SLEW FAST [get_ports {led[3]}]
set_property IOSTANDARD SSTL135 [get_ports {led[3]}]
set_property IN_TERM UNTUNED_SPLIT_40 [get_ports {out_a}]
set_property SLEW FAST [get_ports {out_a}]
set_property IOSTANDARD SSTL135 [get_ports {out_a}]
set_property SLEW FAST [get_ports {out_b[0]}]
set_property IOSTANDARD SSTL135 [get_ports {out_b[0]}]
set_property SLEW FAST [get_ports {out_b[1]}]
set_property IOSTANDARD SSTL135 [get_ports {out_b[1]}]
</pre>

Yosys processes it as follows:
<pre>
Found port led[0]
Setting parameter \DRIVE to value I12 on cell \OBUF_6 

Found port led[1]
Setting parameter \IN_TERM to value UNTUNED_SPLIT_40 on cell \OBUF_7 

Found port led[1]
Setting parameter \SLEW to value FAST on cell \OBUF_7 

Found port led[1]
Setting parameter \IOSTANDARD to value SSTL135 on cell \OBUF_7 

Found port led[2]
Setting parameter \SLEW to value FAST on cell \bottom_inst.OBUF_9 

Found port led[2]
Setting parameter \IOSTANDARD to value SSTL135 on cell \bottom_inst.OBUF_9 

Found port led[3]
Setting parameter \SLEW to value FAST on cell \bottom_intermediate_inst.OBUF_8 

Found port led[3]
Setting parameter \IOSTANDARD to value SSTL135 on cell \bottom_intermediate_inst.OBUF_8 

Found port out_a
Setting parameter \IN_TERM to value UNTUNED_SPLIT_40 on cell \OBUF_OUT 

Found port out_a
Setting parameter \SLEW to value FAST on cell \OBUF_OUT 

Found port out_a
Setting parameter \IOSTANDARD to value SSTL135 on cell \OBUF_OUT 

Found port out_b[0]
Setting parameter \SLEW to value FAST on cell \bottom_inst.OBUF_10 

Found port out_b[0]
Setting parameter \IOSTANDARD to value SSTL135 on cell \bottom_inst.OBUF_10 

Found port out_b[1]
Setting parameter \SLEW to value FAST on cell \bottom_inst.OBUF_11 

Found port out_b[1]
Setting parameter \IOSTANDARD to value SSTL135 on cell \bottom_inst.OBUF_11 
</pre>

The test is run by Travis to make sure subsequent commits don't break anything there.


mithro:
The verilog file in the comment should be a test.

xdc-plugin/xdc.cc
```diff
@@ -98,15 +98,15 @@ struct GetPorts : public Pass {
...
+		if (!sscanf(port_name.c_str(), "%[^[][%d]", port, &bit)) {
+			log_error("Couldn't find port %s\n", port_name.c_str());
 		}
+		std::string port_signal(port);
 
 		RTLIL::IdString port_id(RTLIL::escape_id(port_signal));
 		if (auto wire = top_module->wire(port_id)) {
-			if (bit >= wire->start_offset && bit < wire->start_offset + wire->width) {
-				if (isInputPort(wire) || isOutputPort(wire)) {
+			if (isInputPort(wire) || isOutputPort(wire)) {
```
mkurc-ant: It should not hit us but what about "isInoutPort(wire)"?

xdc-plugin/xdc.cc
```diff
@@ -236,21 +236,37 @@ struct SetProperty : public Pass {
...
+		int port_bit = port_signal.second;
+
+		RTLIL::Wire* wire = design->top_module()->wire(RTLIL::escape_id(port));
+		if (wire == nullptr) {
+			log_error("Couldn't find port %s\n", port_name.c_str());
 		}
-		RTLIL::Module* top_module = design->top_module();
-		RTLIL::IdString port_id(RTLIL::escape_id(port_name));
+
+		if (!isInputPort(wire) && !isOutputPort(wire)) {
```
mkurc-ant: Should handle inouts

tmichalak: I will have to check, but I am pretty sure that for an inout both checks will return true, thus the condition will not be met and the error will not be raised.

xdc-plugin/xdc.cc
tmichalak: Inout wire will return true for both checks so I guess we are good here, but I will double check and update the test accordingly.

mithro:
@hzeller, could you review please?

xdc-plugin/xdc.cc
tmichalak: I checked and inout are handled fine. I added an inout to the test case.

xdc-plugin/xdc.cc
tmichalak: Inout are handled correctly. Added one to the test case.

xdc-plugin/tests/xc7a35tcsg324-1.json
```diff
@@ -0,0 +1,459 @@
+{
```
litghost: Should we make this a smaller file with only what is required?

litghost:
Is having a golden JSON (xdc-plugin/tests/counter_golden.json) the right way to write the test?

.travis.yml
```diff
@@ -0,0 +1,49 @@
+sudo: false
```
litghost: Can you add the travis stuff in a seperate PR?  It seems weird to have it here?

xdc-plugin/xdc.cc
```diff
@@ -268,13 +284,69 @@ struct SetProperty : public Pass {
...
+						return;
+					}
+				}
+			}
+		}
+	}
+
+	// Extract signal name and port bit information from port name
+	std::pair<std::string, int> extract_signal(const std::string& port_name) {
+		char port[128];
```
litghost: It seems weird to be mixing fixed length 128 byte stack strings with std::string.  Can you explain the reasoning why 128 characters is enough?  Why use a stack string over an allocated std::string?

xdc-plugin/tests/xc7a35tcsg324-1.json
tmichalak: We could.

tmichalak:
> Is having a golden JSON (xdc-plugin/tests/counter_golden.json) the right way to write the test

This is the first thing that comes to mind when aiming at avoiding regression. Note that the golden and result files are processed before being compared. One option would be to store only the post processed version as golden reference. Any other suggestions?

litghost:
> > Is having a golden JSON (xdc-plugin/tests/counter_golden.json) the right way to write the test
> 
> This is the first thing that comes to mind when aiming at avoiding regression. Note that the golden and result files are processed before being compared. One option would be to store only the post processed version as golden reference. Any other suggestions?

But golden tests are fragile (e.g. small perturbances result in a meaningless failure).  So my question is what were you trying to test with the diff test?  With the XDC plugin specifically, we only care about the attributes getting bound, so why not test that?

In particular, a diff test of a 10-100 line file is probably okay.  Your golden file is 46k lines long.  Do we *really* care that all 46k lines are exactly the same?

.travis.yml
tmichalak: Sure

tmichalak:
Oh, oh, I totally agree that is why we don't conpare all of them. The test runs a python script that takes the resulting and golden json files and only compares the relevant part, otherwise the test would always fail only due to differences in the path or yosys version.

litghost:
> Oh, oh, I totally agree that is why we don't conpare all of them. The test runs a python script that takes the resulting and golden json files and only compares the relevant part, otherwise the test would always fail only due to differences in the path or yosys version.

Then why are you checking in a 46k line JSON file?

tmichalak:
So one solution would simply be simply extracting the relevant part and storing that as golden reference.

litghost:
> So one solution would simply be simply extracting the relevant part and storing that as golden reference.

I'm fine with that.  There should already be a way to update the golden output, and this can be tacked onto that.

xdc-plugin/xdc.cc
tmichalak: This was because I used sscanf. Otherwise I would go with allocated std::string. I guess using stringstream could be an alternative here.

litghost: You can actually use std::string with sscanf, and because you know the size of the input string, you can safely allocate the maximum extract of port_name.size(), etc.

xdc-plugin/tests/xc7a35tcsg324-1.json
tmichalak: Create PR #5. However it depends on this PR.

.travis.yml
tmichalak: Created PR #5. However it depends on this PR.

xdc-plugin/xdc.cc

.travis.yml
litghost: Why does it depend on this PR?  That seems surprising?

xdc-plugin/tests/counter_golden.json
```diff
@@ -0,0 +1 @@
+{"OBUF_7": {"IN_TERM": "UNTUNED_SPLIT_40", "SLEW": "FAST", "IOSTANDARD": "SSTL135"}, "OBUF_6": {"IOSTANDARD": "LVCMOS33", "DRIVE": "I12", "SLEW": "SLOW"}, "bottom_inst.OBUF_9": {"IOSTANDARD": "SSTL135", "SLEW": "FAST"}, "OBUF_OUT": {"IN_TERM": "UNTUNED_SPLIT_40", "SLEW": "FAST", "IOSTANDARD": "SSTL135"}, "bottom_inst.OBUF_11": {"IOSTANDARD": "SSTL135", "SLEW": "FAST"}, "bottom_inst.OBUF_10": {"IOSTANDARD": "SSTL135", "SLEW": "FAST"}, "bottom_intermediate_inst.OBUF_8": {"IOSTANDARD": "SSTL135", "SLEW": "FAST"}}
```
litghost: Please use "indent=2" on the `json.dumps` to enable better readability.

xdc-plugin/xdc.cc
```diff
@@ -96,17 +96,19 @@ struct GetPorts : public Pass {
 		}
 		// TODO handle more than one port
 		port_name = args.at(1);
-		char port[128];
+		std::string port_str;
+		char* port = const_cast<char*>(port_str.c_str());
```
litghost: If you are using a const cast with the STL library you know you are doing it wrong.  There are actually two bugs here:
 - You should be using `.data()` rather than `.c_str()` to get access to a mutable string
 - When you call `.reserve`, the `std::string` may reallocate, causing the pointer in `port` to become freed memory.

Both errors can result in memory safety violations and must be fixed.

tmichalak: I wouldn't call `.data()` mutable - http://www.cplusplus.com/reference/string/string/data/ and I am not sure there is any difference between `c_str()` and `data()` in the current standard.

xdc-plugin/tests/counter_golden.json
tmichalak: Done

xdc-plugin/xdc.cc
tmichalak: Ok, I state corrected. Seems that since C++17 there is in fact a non-const `.data()`

tmichalak: However, we build the plugins with the same cxx_flags as yosys which currently uses c++11 so the non-const `.data()` is not there. Do we want to enable `-std=c++17` for the plugins?

tmichalak: Regarding `.reserve` is [this](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/4/commits/d474a7659304b2a7df88810cd6f50fe64726e769#diff-bf6d652c14576b97319c59ebd0933b30R99) acceptable?

litghost: If we have to support C++11, then `&port[0]` is acceptable and legal.  Const casting an STL container is almost never legal.

tmichalak: Fixed

xdc-plugin/tests/counter_golden.json
```diff
@@ -0,0 +1,33 @@
+{
+  "OBUF_6": {
+    "DRIVE": "I12",
```
litghost: Shouldn't drive be a number (e.g. "DRIVE": 12)?

xdc-plugin/tests/counter_golden.json
```diff
@@ -0,0 +1,33 @@
...
+    "IOSTANDARD": "LVCMOS33",
+    "SLEW": "SLOW"
+  },
+  "OBUF_7": {
+    "IN_TERM": "UNTUNED_SPLIT_40",
```
litghost: Can you explain this one?  Based on my reading of the verilog (https://github.com/SymbiFlow/yosys-symbiflow-plugins/blob/bb93b7368f885447ee690c44383e4aa13aa0840a/xdc-plugin/tests/counter.v#L26), this is wrong?

xdc-plugin/tests/counter_golden.json
litghost: Ah, I see the problem.  The XDC is wrong is this case, https://github.com/SymbiFlow/yosys-symbiflow-plugins/blob/bb93b7368f885447ee690c44383e4aa13aa0840a/xdc-plugin/tests/counter.xdc#L2 should be `set_property DRIVE 12` not `set_property DRIVE I12`

xdc-plugin/tests/counter_golden.json
litghost: I figured it out, the XDC is overridding the verilog.

xdc-plugin/tests/counter.xdc
```diff
@@ -0,0 +1,22 @@
+#set_property LOC R2 [get_ports led]
+set_property DRIVE I12 [get_ports {led[0]}]
```
litghost: Vary the SLEW/IN_TERM/IOSTANDARD to have confidence that the code is not crossing wires

xdc-plugin/xdc.cc
```diff
@@ -96,17 +96,16 @@ struct GetPorts : public Pass {
...
 		int bit(0);
-		std::string port_signal(port_name);
-		if (sscanf(port_name.c_str(), "%[^[][%d]", port, &bit) == 2) {
-			port_signal = std::string(port);
+		if (!sscanf(port_name.c_str(), "%[^[][%d]", &port_str[0], &bit)) {
+			log_error("Couldn't find port %s\n", port_name.c_str());
 		}
 
-		RTLIL::IdString port_id(RTLIL::escape_id(port_signal));
+		RTLIL::IdString port_id(RTLIL::escape_id(port_str));
```
litghost: You might want to call `port_str.c_str()` so that `escape_id` recomputes the string length.  I don't know of `escape_id` will blindly rely on the std::string::size or not.

xdc-plugin/xdc.cc
```diff
@@ -268,13 +283,69 @@ struct SetProperty : public Pass {
...
+			}
+		}
+	}
+
+	// Extract signal name and port bit information from port name
+	std::pair<std::string, int> extract_signal(const std::string& port_name) {
+		int port_bit(0);
+		std::string port_str(port_name.size(), '\0');
+		sscanf(port_name.c_str(), "%[^[][%d]", &port_str[0], &port_bit);
+		return std::make_pair(port_str, port_bit);
```
litghost: Consider resizing port_str down to "strlen(port_str.c_str())"?

xdc-plugin/tests/counter_golden.json
tmichalak: Sure, in the [xdc file](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/4/files#diff-08bea87e5e93e125875e7fa0dcc142e4R3) the IOSTANDARD gets overriden by setting the IN_TERM, SLEW and IOSTANDARD on led[0] which traverses through two intenal wires to OBUF_7.

xdc-plugin/tests/counter_golden.json
tmichalak: Yepp, fixed.

xdc-plugin/xdc.cc
tmichalak: Fine, that's done.

xdc-plugin/xdc.cc
tmichalak: I have added `port_str.resize(strlen(port_str.c_str()));` one line above, so I guess I don't need one more `port_str.c_str()`.

xdc-plugin/tests/counter.xdc
tmichalak: Done

xdc-plugin/xdc.cc
litghost: Looks fine.

PR #5, opened 2020-01-20
Add Travis scripts to test the plugins 

tmichalak:
This PR adds Travis CI for testing of the plugins.

### DIFF
--- a/.travis/build-and-test.sh
+++ b/.travis/build-and-test.sh
@@ -1,0 +1,61 @@
+#! /bin/bash
+
+set -e
+
+source .travis/common.sh
+
+##########################################################################
+
+echo
+echo 'Configuring...' && echo -en 'travis_fold:start:script.configure\\r'
+echo
+
+if [ "$CONFIG" = "gcc" ]; then
+	echo "Configuring for gcc."
+	make config-gcc
+elif [ "$CONFIG" = "clang" ]; then
+	echo "Configuring for clang."
+	make config-clang
+fi
+
+echo
+echo -en 'travis_fold:end:script.configure\\r'
+echo
+
+##########################################################################
+
+echo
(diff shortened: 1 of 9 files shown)


mithro:
```
Building plugins..
make: *** No rule to make target 'plugins'.  Stop.
The command "./.travis/build-and-test.sh" exited with 2.
```

tmichalak:
Depends on #4. I separated the Travis chages from #4 PR as @litghost requested. However I see how this is a chicken-egg problem, because we want to make sure #4 is ok using Travis. 
I temporarily pushed commits from #4 here to show that Travis works. Once #4 is merged I will rebase and this PR will only have the Travis commit.

litghost:
> Depends on #4. I separated the Travis chages from #4 PR as @litghost requested. However I see how this is a chicken-egg problem, because we want to make sure #4 is ok using Travis.
> I temporarily pushed commits from #4 here to show that Travis works. Once #4 is merged I will rebase and this PR will only have the Travis commit.

What I'd like is a simple PR that sets up an empty test that passes, we merge that, and then #4 adds the test.  Can we do that?

tmichalak:
> What I'd like is a simple PR that sets up an empty test that passes, we merge that, and then #4 adds the test. Can we do that?

I added some dummy test and extracted the test Makefiles from #4.

.travis/build-and-test.sh
```diff
@@ -0,0 +1,61 @@
+#! /bin/bash
+
+set -e
+
+source .travis/common.sh
+
+##########################################################################
+
+echo
```
mithro: Can you create a new issue to replace these with calls to replace these echos with things like https://github.com/litex-hub/litex-conda-packages/blob/9b201004a699af707afc0745e57a06c223e4f545/.travis/common.sh#L32-L43 ?

PR #9, opened 2020-02-21
Fix OOB error in XDC plugin.

litghost:


mithro:
Should there potentially be a test for this?

litghost:
> Should there potentially be a test for this?

I'm not sure how to control how Yosys splits signals.  I'll take a look.

PR #14, opened 2020-05-12
XDC: Fix port index parsing in get_ports command

### LINKED ISSUE #13: XDC plugin doesn't understand port indexes properly
It's a TCL thing. When a port is retrieved using eg. `[get_ports A[10]]` the TCL parser treats the port index in square brackets as another TCL command and fails there. This prevents us from using stock XDC files. A workaround for that is to specify the port like `[get_ports {A[10]}]`. 

A fix may be doing some kind of pre-processing of XDCs prior to feeding them to the Yosys' TCL interpreter.

tmichalak:
This PR fixes #13 by adding a TCL rename proc during `get_ports` command registration.


xdc-plugin/tests/counter.xdc
```diff
@@ -1,27 +1,27 @@
...
+set_property SLEW FAST [get_ports out_a]
 set_property IOSTANDARD LVCMOS33 [get_ports {out_a}]
 #bottom_inst.OBUF_10
 set_property SLEW SLOW [get_ports {out_b[0]}]
 set_property IOSTANDARD LVCMOS18 [get_ports {out_b[0]}]
 #bottom_inst.OBUF_11
 set_property DRIVE 4 [get_ports {out_b[1]}]
 set_property SLEW FAST [get_ports {out_b[1]}]
-set_property IOSTANDARD LVCMOS25 [get_ports {out_b[1]}]
+set_property IOSTANDARD LVCMOS25 [get_ports out_b[1]]
```
mithro, line 18: Why does this line change but not line 21?

tmichalak: This is because `get_ports out_b[1]` should be equivalent to `get_ports {out_b[1]}` hence the test contains both versions in random places.

tmichalak:
@litghost so what you're saying is that we should stay with what we had before?

litghost:
> @litghost so what you're saying is that we should stay with what we had before?

What evidence do we have that the current implementation is incorrect?

tmichalak:
> What evidence do we have that the current implementation is incorrect?

The current implementation was correct so this PR is more an enhancement for convenience of use reported in issue #13. In my implementation both versions `get_ports a[0]` and `get_ports {a[0]}` are equivalent. It's similar to the tcl shell in Vivado.

mithro:
@tmichalak -- You should have separate tests which are clearly says what they are testing.

xdc-plugin/tests/port_indexes.xdc
```diff
@@ -0,0 +1,30 @@
...
+#bottom_inst.OBUF_11
+set_property DRIVE 4 [get_ports out_b[1]]
+set_property SLEW FAST [get_ports out_b[1]]
+set_property IOSTANDARD LVCMOS25 [get_ports out_b[1]]
+#bottom_inst.OBUF_9
+set_property SLEW FAST [get_ports led[2]]
+set_property IOSTANDARD DIFF_SSTL135 [get_ports led[2]]
+#bottom_intermediate_inst.OBUF_8
+set_property DRIVE 16 [get_ports led[3]]
+set_property IOSTANDARD SSTL135 [get_ports led[3]]
```
litghost, line 24: Should this test both with and without quote, with a comment to each affect?

xdc-plugin/xdc.cc
```diff
@@ -385,9 +384,25 @@ struct ReadXdc : public Frontend {
 		extra_args(f, filename, args, argidx);
 		std::string content{std::istreambuf_iterator<char>(*f), std::istreambuf_iterator<char>()};
 		log("%s\n", content.c_str());
+
+		// According to page 6 of UG903 XDC is tcl, hence quuoting of bracketed numbers,
+		// such as bus indexes, is required. However, it's quite common for EDA tools
+		// to allow for correct processing of the bracketed numbers without quoting.
```
litghost, line 390: > correct processing

I object to this wording.  I would rephrase it.

xdc-plugin/xdc.cc
```diff
@@ -385,9 +384,25 @@ struct ReadXdc : public Frontend {
...
+		// Possible TCL implementations of such a feature include registering a TCL command
```
litghost, line 391: This needs explicit examples.  E.g.

> For example, tcl would typical expand:
>
>> signal[5]
>
> as the string "signal" concatenated with the function call "5" with no arguments.

Then explain in detail how the modification to the unknown function handles this.

xdc-plugin/tests/port_indexes.xdc
tmichalak: The `port_indexes` test is basically the counter test (with curly braces), but without curly braces.

xdc-plugin/xdc.cc

xdc-plugin/xdc.cc
tmichalak: Done

litghost: Good, thanks!

xdc-plugin/xdc.cc
litghost: Good, thanks!

xdc-plugin/tests/port_indexes.tcl
```diff
@@ -0,0 +1,25 @@
...
+# -flatten is used to ensure that the output eblif has only one module.
+# Some of symbiflow expects eblifs with only one module.
+synth_xilinx -vpr -flatten -abc9 -nosrl -noclkbuf -nodsp
+
+if {[info procs unknown] != ""} {
+	rename unknown ""
+}
+proc unknown args {puts "'unknown' proc command handler"}
+if {[catch {invalid command}]} {
+	error "Command should be handled by the 'unknown' proc"
```
litghost, line 15: You need to validate that "'unknown' proc command handler"" gets output twice?

tmichalak: Done

PR #15, opened 2020-06-04
Add format target and format cpp files

tmichalak:
Use ClangFormat to format the cpp files and add a format target to the Makefile.


mithro:
@tgorochowik I like this patch! However, what format style is this using?

tmichalak:
@mithro I forgot to add it. It is the same as in X-Ray.

mithro:
@tmichalak -- Should probably use the Yosys config.

mithro:
https://github.com/YosysHQ/yosys/blob/master/.clang-format

xdc-plugin/xdc.cc
```diff
@@ -81,16 +73,18 @@ struct GetPorts : public Pass {
 		log("\n");
 		log("Get matching ports\n");
 		log("\n");
-		log("Print the output to stdout too. This is useful when all Yosys is executed\n");
+		log("Print the output to stdout too. This is useful when all "
```
mithro, line 76: What the heck is going on here?

litghost:
This needs a rebase?

.clang-format
```diff
@@ -0,0 +1,12 @@
+# Default Linux style
+BasedOnStyle: LLVM
+IndentWidth: 8
+UseTab: Always
```
litghost, line 4: Please use:
```
IndentWidth: 4
TabWidth: 4
UseTab: Never
```

tmichalak: Fixed

PR #26, opened 2020-07-08
Get param plugin

tmichalak:
This PR adds the get_param command, which along with the existing setparam command can be used for manipulating the cell parameter values using TCL, e.g `expr`


tmichalak:
It's still wip as the parameter value returned by the tcl interpreter is garbage.
The get_param command retrieves the value correctly.
```
get_param CLKIN1_PERIOD top/PLLE2_ADV
setparam -set CLKIN1_PERIOD {"20.0"} top/PLLE2_ADV
get_param CLKIN1_PERIOD top/PLLE2_ADV
```
returns
```
CLKIN1_PERIOD:10.000000
CLKIN1_PERIOD:20.0
```

litghost:
> CLKIN1_PERIOD:10.000000
> CLKIN1_PERIOD:20.0

Why would you return the parameter name too?  Why not just return the value?

param-plugin/params.cc
```diff
@@ -0,0 +1,76 @@
...
+			for (auto cell : module->selected_cells()) {
+				auto params = cell->parameters;
+				auto it = params.find(RTLIL::IdString(RTLIL::escape_id(param)));
+				if (it != params.end()) {
+					value = it->second.decode_string();
+				}
+			}
+		}
+		Tcl_Interp *interp = yosys_get_tcl_interp();
+		Tcl_SetResult(interp, const_cast<char*>(value.c_str()), NULL);
```
litghost, line 69: This is not legal C++, because the malloc'd array in value leaves scope.

litghost: From the documentation on `Tcl_SetResult`:

```
Tcl_SetResult arranges for result to be the result for the current Tcl command in interp, replacing any existing result. The freeProc argument specifies how to manage the storage for the result argument; it is discussed in the section THE TCL_FREEPROC ARGUMENT TO TCL_SETRESULT below. If result is NULL, then freeProc is ignored and Tcl_SetResult re-initializes interp's result to point to an empty string.
```

You need to provide a `freeProc` per the documentation to free the string once the interperator is done with it.

litghost: Reading further, it looks like you need to allocation a new Tcl memory space using https://www.tcl.tk/man/tcl8.5/TclLib/Alloc.htm and set freeProc = `TCL_DYNAMIC`.

tmichalak:
> Why would you return the parameter name too? Why not just return the value?

The function's output and return value are separate things, so why not print out both?
However, currently I do this for debugging purposes.

litghost:
> > Why would you return the parameter name too? Why not just return the value?
> 
> The function's output and return value are separate things, so why not print out both?
> However, currently I do this for debugging purposes.

Sorry, I interpreted that output as the return value.  Once you finish debugging the TCL return value, I don't think the print is needed.

param-plugin/params.cc
tmichalak: Thanks for the info. I remembered there is something to do with memory to make the Tcl function in C++ work, hence this is still WIP.

tmichalak: Fixed.

tmichalak:
Still need to work on the test, but the command works.

tmichalak:
The test has been updated and the plugins tests have been added to the list of all tests. Ready for review, provided the CI is green.

params-plugin/params.cc
```diff
@@ -0,0 +1,82 @@
...
+		if (args.size() == 1) {
+			log_error("Incorrect number of arguments");
+		}
+
+		std::string param(args.at(1));
+		std::string value;
+		extra_args(args, 2, design);
+
+		for (auto module : design->selected_modules()) {
+			for (auto cell : module->selected_cells()) {
```
litghost, line 60: Should we allow selection 0 or >1 elements?  In the >1 elements case, this code will only output the last property.

Another option is to return a TCL list if >1 elements

mithro:
Random question -- Is it `set_param` and `get_param` or `setparam` and `getparam` -- hopefully not `set_param` and `getparam`?

litghost:
Yosys provides `setparam`.  Looks like the current plugin is `get_param`.  @tmichalak, any reason we cannot match yosys's convention here?

tmichalak:
> any reason we cannot match yosys's convention here?

No other reason than that I think setparam is worse than set_param as in other EDA tools, Yosys included in case of other commands. But since we won't be able to change that it makes sense to make our command consistent, which in this case means equally bad.

params-plugin/params.cc
```diff
@@ -0,0 +1,82 @@
...
+				auto params = cell->parameters;
+				auto it = params.find(RTLIL::IdString(RTLIL::escape_id(param)));
```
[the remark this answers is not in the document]
tmichalak: Fixed.

tmichalak:
> Why is params-plugin/tests/xilinx/cells_xtra.v and params-plugin/tests/techmaps/cells_map.v being added? I don't believe it makes sense to add 38k lines of verilog to test this plugin.

Right, they are not needed, since they are supposed to be taken from the Yosys's share directory. I removed them.

params-plugin/params.cc
tmichalak: I guess >1 elements makes sense because it's how get_property work in Vivado for example.

tmichalak: Fixed.

params-plugin/tests/pll/pll.tcl
```diff
@@ -0,0 +1,32 @@
+yosys -import
+plugin -i xdc
+plugin -i params
+#Import the commands from the plugins to the tcl interpreter
+yosys -import
+
+read_verilog -specify -lib -D_EXPLICIT_CARRY +/xilinx/cells_sim.v
+read_verilog -lib +/xilinx/cells_xtra.v
+hierarchy -check -auto-top
+set phase [getparam CLKOUT2_PHASE top/PLLE2_ADV_0 top/PLLE2_ADV]
```
litghost, line 10: This test doesn't actually verify that phase returns:
 - Correct number of values
 - That the values are correct?

params-plugin/Makefile
```diff
@@ -0,0 +1,25 @@
...
+OBJS = params.o
+
+params.so: $(OBJS)
+	$(CXX) $(CXXFLAGS) $(LDFLAGS) -shared -o $@ $^ $(LDLIBS)
+
+install_plugin: params.so
+	mkdir -p $(PLUGINS_DIR)
+	cp $< $(PLUGINS_DIR)/$<
+
+test: install_plugin
```
litghost, line 16: Don't install when testing

params-plugin/tests/pll/pll.tcl
```diff
@@ -0,0 +1,48 @@
...
+} else {
+	puts $fp "FAIL"
+}
+setparam -set CLKOUT2_PHASE [expr [lindex $phase 0] * 1000] top/PLLE2_ADV
+set phase [getparam CLKOUT2_PHASE top/PLLE2_ADV_0 top/PLLE2_ADV]
+puts -nonewline $fp "Phase after: "
+if {$phase == [list 90000 70]} {
+	puts $fp "PASS"
+} else {
+	puts $fp "FAIL"
```
litghost, line 27: Might want to emit more data on failure?

params-plugin/Makefile
tmichalak: Fixed.

params-plugin/tests/pll/pll.tcl
tmichalak: Sure, fixed.

PR #27, opened 2020-07-17
SDC plugin

tmichalak:
In this PR the following commands have been implemented:
* read_sdc
* create_clock
* get_clocks
* write_sdc

To Do:
- [x] SDC import
- [x] Natural propagation pass - through wires
- [x] Propagation across buffers (IBUF, BUFG)
- [x] Propagation across PLLs
- [ ] Propagation across MMCMs (Optional)
- [x] SDC export
- [x] Cleanup and fixing tests


sdc-plugin/buffers.cc
```diff
@@ -0,0 +1,24 @@
...
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+#include "buffers.h"
+
+const std::vector<std::string> Pll::inputs = {"CLKIN1", "CLKIN2"};
+const std::vector<std::string> Pll::outputs = {"CLKOUT0", "CLKOUT1", "CLKOUT2",
+                                               "CLKOUT3", "CLKOUT4", "CLKOUT5"};
+const float Pll::delay = 1;
```
litghost, line 23: Why is `delay = 1`?  Should just be `delay = 0`?

sdc-plugin/tests/counter/counter.input.sdc
```diff
@@ -0,0 +1,2 @@
+create_clock -period 10.0 -waveform {0.000 5.000} clk_int_1
+create_clock -period 10.0 -name clk -waveform {0.000 5.000} clk clk2
```
litghost, line 2: Should add a phase offset example?

sdc-plugin/buffers.cc
tmichalak: It's just there for the first cut. Unless there will be no delay on the PLL at all...

tmichalak:
>  Have you tested this with some examples from symbiflow yet?

I am working on integrating that with arch-defs locally and want to test it on mini_ddr and base litex. I already spotted some small bugs which I will work on

> Also, what happens if the user specifies the clock constraints in the XDC file rather than a SDC file? Will this work too?

Yes, it will, since the SDC plugin basically implements a create_clock command and it doesn't matter whether it gets read by `read_xdc` , `read_sdc` or `source` commands. However the `read_sdc` command doesn't require the `-part` switch for example. Moreover, at some point we might want to add some arguments to `read_sdc` for example. If not, then a `read_sdc` command in the script is more descriptive.

litghost:
> > Also, what happens if the user specifies the clock constraints in the XDC file rather than a SDC file? Will this work too?
> 
> Yes, it will, since the SDC plugin basically implements a create_clock command and it doesn't matter whether it gets read by `read_xdc` , `read_sdc` or `source` commands. However the `read_sdc` command doesn't require the `-part` switch for example. Moreover, at some point we might want to add some arguments to `read_sdc` for example. If not, then a `read_sdc` command in the script is more descriptive.

Awesome!

litghost:
> > Have you tested this with some examples from symbiflow yet?
> 
> I am working on integrating that with arch-defs locally and want to test it on mini_ddr and base litex. I already spotted some small bugs which I will work on

Should we merge this as is, or wait till you finish debugging locally?

tmichalak:
> Should we merge this as is, or wait till you finish debugging locally?

I can do the debugging without merging, so I would wait with it until I can confirm that mini_ddr and base litex are working locally (should be by the end of this week). Then we can merge and put up the arch-defs integration PR. If we find bugs past that point we can just open issues on yosys-symbiflow-plugins.

tmichalak:
@litghost:
I successfully verified that the minilitex_ddr design works on hardware with the SDC produced with the plugin.
The branch that integrates the plugin into arch-defs is [here](https://github.com/antmicro/symbiflow-arch-defs/tree/sdc_plugin_integration). I will create a PR once the plugin is upstream so I can update the plugin version in arch-defs accordingly

sdc-plugin/buffers.h
```diff
@@ -0,0 +1,93 @@
...
+struct Buffer {
+    Buffer(float delay, const std::string& name, const std::string& output)
+        : delay(delay), name(name), output(output) {}
+    float delay;
+    std::string name;
+    std::string output;
+};
+
+struct IBuf : Buffer {
+    IBuf() : Buffer(1, "IBUF", "O"){};
```
litghost, line 39: Still have delay as `1`, which I assume was a placeholder?

sdc-plugin/tests/pll/pll.tcl
```diff
@@ -0,0 +1,23 @@
...
+hierarchy -check -auto-top
+
+# Start flow after library reading
+synth_xilinx -vpr -flatten -abc9 -nosrl -nodsp -iopad -run prepare:check
+
+# Read the design timing constraints
+read_sdc $::env(INPUT_SDC_FILE)
+
+# Propagate the clocks
+propagate_clocks
```
litghost, line 18: Why does it need to be called 3 times?

sdc-plugin/buffers.h
tmichalak: So we don't want any delay through IBUF. This was meant to be a placeholder for any value we want to have there in the future.

sdc-plugin/tests/pll/pll.tcl
tmichalak: Explained it [here](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1647#discussion_r476605300)

sdc-plugin/buffers.h
litghost: There is a delay, but 1 ns is not the right number

tmichalak: I will change it to 0 for now? We can create an issue to make sure we get the right numbers in there.

mithro:
@tmichalak - Just wanted to confirm that you saw the [spreadsheet I created here](https://docs.google.com/spreadsheets/d/1G-E2Dq8YG4g9Z6mTygpumwlI_vNlFUQinc9gMgePfec/edit#gid=0). Can you make sure to update the spreadsheet with the status of supported commands as this gets merged?

mithro:
@tmichalak - How hard would this be to split this plugin into two parts?

 - (a) 'sdc' -- The part which reads SDC into the Yosys attributes.
 - (b) 'propagate-constraints' -- The part which propagates the constraints in the Yosys attributes across things like clock buffers and similar.

The idea being that the propagation is *independent* of how the constraints got into the Yosys attributes. IE We might end up wanting a `read_pcf` or `read_other_timing_constraint_format` plugins which pull data into the same Yosys attributes.

tmichalak:
@mithro 
>  Just wanted to confirm that you saw the spreadsheet I created here. Can you make sure to update the spreadsheet with the status of supported commands as this gets merged?

Yes, I saw it and I will fill it out before we get this merged.

> How hard would this be to split this plugin into two parts?

I think it's already divided the way you described in the current implementation, i.e. another command, like read_pcf would populate the same structure with clock information like sdc. The propagation steps just extend the clock information based on the cells found in the design. It's already separate thanks to the visitor design pattern I used in the implementation.

mithro:
> > How hard would this be to split this plugin into two parts?
> 
> I think it's already divided the way you described in the current implementation, i.e. another command, like read_pcf would populate the same structure with clock information like sdc. The propagation steps just extend the clock information based on the cells found in the design. It's already separate thanks to the visitor design pattern I used in the implementation.

I was suggesting going even further and putting them into totally separate plugins.

tmichalak:
> I was suggesting going even further and putting them into totally separate plugins.

It depends on what you would like to have in each plugin. I guess one should expose commands like `read_sdc, read_pcf, write_sdc, create_clock` and other sdc commands and another plugin should have commands like `propagate_clocks -natural|-buffer|-clock_divider`?

mithro:
> > I was suggesting going even further and putting them into totally separate plugins.
> 
> It depends on what you would like to have in each plugin. I guess one should expose commands like `read_sdc, read_pcf, write_sdc, create_clock` and other sdc commands and another plugin should have commands like `propagate_clocks -natural|-buffer|-clock_divider`?

I would say;
 - `sdc` plugin --> `read_sdc`, `write_sdc`
 - `pcf` plugin --> `read_pcf`
 - `clock_tools` plugin --> `propagate_clocks` (and maybe future things like `generate_pll_config` or something?)

I'm unsure what `create_clock` is or what it does?

tmichalak:
> I would say;
>
> sdc plugin --> read_sdc, write_sdc
> pcf plugin --> read_pcf
> clock_tools plugin --> propagate_clocks (and maybe future things like generate_pll_config or something?)
> I'm unsure what create_clock is or what it does?

Makes sense, however sharing an object between plugins is something I would have to think about. Currently, the clock information is not assigned to parameters but kept in a separate Clocks object. This would have to be global in order to work for all plugins. Another option would be to have the information stored in the RTLIL::Wire parameters, but that would require some extra implementation.
The bottom line here is whether we want to implement the SDC support the way you described (in 3 plugins) right away or merge the current implementation and then enhance it in the future.

tmichalak:
@mithro I updated the spreadsheet with the supported SDC commands. Judging from the spreadsheets there is still plenty to be done on the plugins front...

sdc-plugin/buffers.h
tmichalak: Fixed.

tmichalak:
@litghost I tested the mini_ddr litex design generated automatically locally using [this PR](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1657) and I believe it's time for another round of review comments and then merging the main PR, so that we can open new tickets for discovered issues. The plugins repo is becoming richer and will most probably grow even bigger, so we might as well get used to opening issues for some bugs that will show up during integration.

sdc-plugin/clocks.cc
```diff
@@ -0,0 +1,255 @@
...
+}
+
+void Clock::UpdateWires(RTLIL::Wire* wire) {
+    if (std::find(clock_wires_.begin(), clock_wires_.end(), wire) ==
+        clock_wires_.end()) {
+	clock_wires_.push_back(wire);
+    }
+}
+
+void Clock::UpdatePeriod(float period) {
```
litghost, line 237: This is assuming a 50% duty cycle.  Did you check this assumption when reading in the clock constraint?

sdc-plugin/clocks.cc
```diff
@@ -0,0 +1,255 @@
...
+    period_ = period;
+    rising_edge_ = 0;
+    falling_edge_ = period / 2;
+}
+
+void Clock::UpdateWaveform(float rising_edge, float falling_edge) {
+    rising_edge_ = rising_edge;
+    falling_edge_ = falling_edge;
+    assert(falling_edge - rising_edge == period_ / 2);
```
litghost, line 246: This probably needs to be a delta with error, like `((falling_edge - rising_edge) - (period_ / 2)) <= period_*(10*std::numeric_limits<decltype(period_)>.epsilon())`

litghost: This should probably not be an `assert`, as this can be compiled out.  Probably just a `if violated, raise error`?

litghost:
> @litghost I tested the mini_ddr litex design generated automatically locally using [this PR](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1657) and I believe it's time for another round of review comments and then merging the main PR, so that we can open new tickets for discovered issues. The plugins repo is becoming richer and will most probably grow even bigger, so we might as well get used to opening issues for some bugs that will show up during integration.

I have some small comments about the way we are handling waveform, but then I think we are good to merge the first cut at this.

sdc-plugin/clocks.cc
tmichalak: In my opinion we don't a check at all. I mean after all this plugin is a SDC commands "forwarder" which takes the users SDC and creates another one based on this input. The real consumer is VPR and I believe there are checks there to parse and validate these values. I don't think the plugin is the place the perform such checks. Besided non 50% duty cycle is supported in VPR, however according to the documentation it's being treated as 50% anyways.

sdc-plugin/clocks.cc
tmichalak: When I am looking at this now I come to think that we don't really want to change the waveform based on the period assuming 50% duty cycle?

litghost: I actually think this particular bit of code is right, but maybe by accident.  If when creating a clock, you only specify the period, then the duty cycle is assumed to be 50%, which is what this code does.  Then if you also specify the waveform, it updates the rising edge and falling edge times.

sdc-plugin/clocks.cc
litghost: So something to watch out for is if a non-0 deg phase is put through a buffer or PLL, then output clock signal should respect the phase.

To be more concrete:
 - Buffers (like IBUF and BUFG) should output a clock with the same frequency and phase shift equal to the delay through the buffer
 - Clock synth's (like MMCM and PLL) should output a clock with the computed frequency, and a phase shift equal to the phase specified.  For example if the PLL input has a 20 degree phase shift, and the request to the PLL is to add another 45 degree phase shift, the output waveform should have a 65 degree phase shift.

Does that make sense?

tmichalak: It does make sense. The buffers are already implemented this way, for PLLs we currently have the delay set to 0, so on the output we want the phase of the input + the requested shift + delay (equal to 0). Is this correct?

litghost: > It does make sense. The buffers are already implemented this way, for PLLs we currently have the delay set to 0, so on the output we want the phase of the input + the requested shift + delay (equal to 0). Is this correct?

I believe so.  Something to watch for is what happens when the total phase shift is > 180 degrees.

tmichalak:
I rebased the PR and enhanced the clock divider propagation to take the duty cycle and phase parameters into account as well as the shift of the input clock.
We still may want to add a waveform sanity check to AddClock.

sdc-plugin/clocks.cc
tmichalak: Check is done on waveform update.

sdc-plugin/tests/counter/counter.input.sdc
tmichalak: Done

tmichalak:
I believe the initial version of the plugin is finished. IMO we should merge it and address dedicated issues.

PR #29, opened 2020-08-27
Selection plugin: Initial version of selection_to_tcl_list command

tmichalak:
Initial version of the selection_to_tcl_list command.

### DIFF
--- a/selection-plugin/selection.cc
+++ b/selection-plugin/selection.cc
@@ -1,0 +1,91 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
+ *  Copyright (C) 2020  The Symbiflow Authors
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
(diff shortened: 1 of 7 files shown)


tmichalak:
@litghost 
This is just the initial version. I could use some advice on the command's interface that has a chance of being accepted upstream. Maybe Eddie could be helpful here?

selection-plugin/selection.cc
```diff
@@ -0,0 +1,91 @@
...
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("   selection_to_tcl_list selection\n");
+		log("\n");
+		log("Extract the current selection to a Tcl List with selection object names. \n");
+		log("\n");
+	}
+
+	void AddObjectNameToTclList(RTLIL::IdString& module, RTLIL::IdString& object, Tcl_Obj* tcl_list) {
+		std::string name = RTLIL::unescape_id(module) + "/" + RTLIL::unescape_id(object);
```
litghost, line 42: Why `unescape_id` vs `id2cstr`? See https://github.com/YosysHQ/yosys/blob/ede4b10da8fdbdcff900b93c6c723516901483ff/passes/cmds/select.cc#L1444

tmichalak: Because `id2cstr` is to c_str whereas `unescape_id` is to string. Considered `std::string(id2cstr())`.

litghost: As long as they provide the same contents, I'm good.  I was just comparing and noticed the difference.

selection-plugin/tests/counter/counter.golden.txt
```diff
@@ -0,0 +1,3 @@
+1
```
litghost, line 1: This golden looks broken now, what happened?

litghost: I see the problem.  The golden should probably have actually output from `selection_to_tcl_list` and generate a TCL error if the comparision to `-write` fails

tmichalak: Fixed, the reference contains the selection now and the test returns a tcl error upon failure.

PR #35, opened 2020-09-15
Add get_nets command

tmichalak:
```
   get_nets [-quiet] [-filter filter_expression] <net_selection>                                                     │~                                                                                                                       
                                                                                                                     │~                                                                                                                       
Get matching nets                                                                                                    │~                                                                                                                       
                                                                                                                     │~                                                                                                                       
Print the output to stdout too. This is useful when all Yosys is executed.                                            │~                                                                                                                       
                                                                                                                     │~                                                                                                                       
    -filter                                                                                                          │~                                                                                                                       
        Name and value of attribute to be taken into account.                                                        │~                                                                                                                       
        e.g. -filter { attr == "true" }                                                                              │~                                                                                                                       
                                                                                                                     │~                                                                                                                       
    -quiet                                                                                                           │~                                                                                                                       
        Don't print the result of the execution to stdout.                                                           │~                                                                                                                       
                                                                                                                     │~                                                                                                                       
    <net_selection>                                                                                                  │~                                                                                                                       
        Selection of net name. Default are all nets in the design.   
```
The command has been added to the XDC plugin.

### DIFF
--- a/xdc-plugin/xdc.cc
+++ b/xdc-plugin/xdc.cc
@@ -191,7 +191,7 @@ struct GetNets : public Pass {
 		size_t argidx;
 		std::vector<std::pair<std::string, std::string>> filters;
 		bool is_quiet = false;
-		bool is_filter = false;
+		bool has_filter = false;
 
 		// Parse command arguments
 		for (argidx = 1; argidx < args.size(); argidx++) {
@@ -211,21 +211,22 @@ struct GetNets : public Pass {
 			std::regex filter_attr_regex("(\\w+\\s?==\\s?\\w+)([(||)(&&)]*)");
 			std::regex_token_iterator<std::string::iterator> regex_end;
 			std::regex_token_iterator<std::string::iterator> matches(filter_arg.begin(), filter_arg.end(), filter_attr_regex, 1);
+			if (matches == regex_end) {
+			    log_warning("Currently -filter switch supports only a single 'equal(==)' condition expression, the rest will be ignored\n");
+			}
 
 			while (matches != regex_end) {
 			    std::string filter(*matches++);
 			    auto separator = filter.find("==");
 			    if (separator == std::string::npos) {
-				log_cmd_error("Incorrect filter expression: %s %s\n", arg.c_str(), args[argidx].c_str());
+				log_cmd_error("Incorrect filter expression: %s\n", args[argidx].c_str());
 			    }
 			    filters.emplace_back(filter.substr(0, separator), filter.substr(separator + 2));
 			}
-
-			if (size_t filter_cnt = filters.size()) {
-			    if (filter_cnt > 1) {
-				log_warning("Currently -filter switch supports only a single expression, the rest will be ignored\n");
-			    }
-			    is_filter = true;
+			size_t filter_cnt = filters.size();
+			has_filter = filter_cnt > 0;
+			if (filter_cnt > 1) {
+			    log_warning("Currently -filter switch supports only a single 'equal(==)' condition expression, the rest will be ignored\n");
 			}
 			continue;
 		    }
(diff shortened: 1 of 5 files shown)


xdc-plugin/xdc.cc
```diff
@@ -154,6 +154,107 @@ struct GetIOBanks : public Pass {
...
+		for (argidx = 1; argidx < args.size(); argidx++) {
+		    std::string arg = args[argidx];
+		    if (arg == "-quiet") {
+			is_quiet = true;
+			continue;
+		    }
+		    if (arg == "-filter" and argidx + 1 < args.size()) {
+			std::string filter_arg = args[++argidx];
+			filter_arg.erase(std::remove_if(filter_arg.begin(), filter_arg.end(), isspace), filter_arg.end());
+			auto separator = filter_arg.find("==");
```
litghost, line 205: This is a pretty narrow version of "-filter", should detect and generate errors and/or warnings if the limited filter support is exceeded by the call.

tmichalak:
@litghost I am working on extending the filter functionality, but for now I am display a suitable warning message.
We could merge the function with the limited `-filter` support as it's blocking the litex auto-generation PR which is using `get_nets` (without the -filter at all).

xdc-plugin/xdc.cc
tmichalak: Changed the implementation to easily extend for proper filter support. Added a warning in the current narrow version.

xdc-plugin/xdc.cc
```diff
@@ -154,6 +155,127 @@ struct GetIOBanks : public Pass {
...
+			while (matches != regex_end) {
+			    std::string filter(*matches++);
+			    auto separator = filter.find("==");
+			    if (separator == std::string::npos) {
+				log_cmd_error("Incorrect filter expression: %s %s\n", arg.c_str(), args[argidx].c_str());
+			    }
+			    filters.emplace_back(filter.substr(0, separator), filter.substr(separator + 2));
+			}
+
+			if (size_t filter_cnt = filters.size()) {
```
litghost, line 224: This is a pretty strange way to write this.  Why not just write:
```
size_t filter_cnt = filters.size();
has_filter = filter_cnt > 0;
if (filter_cnt > 1) {
  log_warning("Currently -filter switch supports only a single expression, the rest will be ignored\n");
}
```

xdc-plugin/xdc.cc
```diff
@@ -154,6 +155,127 @@ struct GetIOBanks : public Pass {
...
+		    }
+
+		    if (arg == "-filter" and argidx + 1 < args.size()) {
+			std::string filter_arg = args[++argidx];
+
+			// Remove spaces
+			filter_arg.erase(std::remove_if(filter_arg.begin(), filter_arg.end(), isspace), filter_arg.end());
+
+			// Parse filters
+			std::regex filter_attr_regex("(\\w+\\s?==\\s?\\w+)([(||)(&&)]*)");
```
litghost, line 211: This doesn't support the full range of filters?  I believe filter supports `<`, `<=` `=~`, etc.  I'm fine with only implementing `==`, but I'd like to see a warning on valid but unsupported filter expressions.  Right now, if the regex doesn't match, I don't think we emit anything?

Maybe just emit a warning if the regex doesn't consume all input?

tmichalak: We do emit an error about incorrect filter expression if there is no '=='.

tmichalak: Oh, if it doesn't match... Yeah at this point we don't do anything.

tmichalak: Added a warning if other than 'equal' condition expression is used.

xdc-plugin/xdc.cc
tmichalak: Fixed.

tmichalak:
Suprseded by #42

PR #37, opened 2020-09-17
SDC: Handle input clock phase shift on clock generators

tmichalak:
Opening this PR for discussion purposes.

For the initial version of the SDC plugin we had a [discussion](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/27#discussion_r485751648) around phase shift and validation of the waveform.
In the current implementation we have a 1ns delay on BUFGs and clock synth's (like MMCM and PLL) output a clock with the computed frequency, and a phase shift equal to the phase specified. For example if the PLL input has a 20 degree phase shift, and the request to the PLL is to add another 45 degree phase shift, the output waveform will have a 65 degree phase shift.

However, this sometimes leads to sanity checks being raised:
```
`8. Executing Verilog-2005 frontend: /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/env/conda/envs/symbiflow_arch_def_base/bin/../share/yosys//plugins/fasm_extra_modules/BANK.v
ERROR: Phase shift on clock sys4x_dqs_clk exceeds 360 degrees
Rising edge: 3.041667, Falling edge: 5.125000, Clock period:4.166667
```
Not sure how vivado handles these situations though.


mkurc-ant:
On page 66 in https://www.xilinx.com/support/documentation/user_guides/ug472_7Series_Clocking.pdf are details on how the phase detector of a PLL works. Judging from that I conclude that the whole thing behaves as an "integrating regulator" so it can synchronize both phases and frequencies of CLKINn and CLKFB.

I also think that just adding phases is wrong as the frequency of a PLL output may be different than of CLKINn. With that phases on both clocks have different meanings and cannot be added/compared directly.

litghost:
> I also think that just adding phases is wrong as the frequency of a PLL output may be different than of CLKINn. With that phases on both clocks have different meanings and cannot be added/compared directly.

The PLL's output phase should be related to one of the input phases in some way.  I agree that it isn't a straight summation isn't the right answer, but having the output clocks have no relation to the input phase is wrong too?  The output of the VCO in a PLL matches the phase of the input clock.  The VCO output can then be phase shifted as needed.

See page 74 and eq 3-3, 3-4, 3-5.  They are all speaking about shifting phase with respect to the input clock.

litghost:
So I think I understand the problem, and what the solution is.  Let's ignore frequency changes for a minute.  If a 10 ns period, 0 deg 50% wave form enters a PLL that is the following create clock:

```
create_clock -period 10 -waveform {0 5} clk
```

So the clock raises on 0 and 1*period and 2*period, etc.
The clock falls on 5 and 5+1*period, etc.

If we apply a 270 degree phase shift to clk, it will be:

```
create_clock -period 10 -waveform {7.5 2.5} clk
```

So the clock raises at 7.5 and 7.5+1*period and 7.5+2*period, etc.
The failing clock case is interesting.  It could be thought as having either a 12.5 falling edge, or a 2.5 falling edge, they are equivalent.  I believe SDC requires the edge time be modulus the period.  For clarity: `2.5 + 1*period == 12.5 + 0*period`.

So one bug that needs to be fixed is the waveform values need to be modulus the period.

That still leaves the meaning of the phase relationship through a PLL when there is a frequency shift.  I believe I know the answer, but want to think on it more.

litghost:
I've done some work, and I believe I understand how to relate the input clock phase and the output clock phase on the PLL.

If the PLL output phase shift = 0 deg, then the rising edge of the output clock matches the rising edge of the VCO.  When FBCLKOUT phase shift = 0, this implies that the VCO and input clock rising edge are phase matched.

If the PLL output phase shift != 0, the output clock is phase shifted per it's output frequency, with respect to the output phase shift = 0 case.

So for example:

If the input clock is 100 MHz, with a phase shift of 90 degree, that implies the following create clock:

```
create_clock -period 10 -waveform {2.5 7.5} clk
```

An output clock of 200 MHz with an output phase shift of 0 should have a rising edge at 2.5 ns, so the create clock is:
```
create_clock -period 5 -waveform {2.5 0} clkout
```

An output clock of 200 MHz with an output phase shift of 90 deg would have a create clock of:
```
create_clock -period 5 -waveform {3.75 1.25} clkout
```

So I believe this is how to apply input phase to output phase:

1. Compute the input clocks rising edge time
2. Create a non-phase shifted output at the new frequency that shares that rising edge, e.g. the remainder of the rising edge time and the new output clock period
3. If the PLL is apply a phase shift, shift the non-shifted waveform by `(phase shift / 360) * period ns`.  If either the rising or falling edge are greater than period, then subtract one period from that wave form.
3a.  For output clocks with duty cycle != 50%, compute the falling edge time from the `rising edge time + duty cycle * period`

@mkurc-ant / @tmichalak Do you disagree with anything here?  I'm going to do some testing to figure out how to accommodate feedback phase shifts.

litghost:
Okay, looks like `CLKFBOUT_PHASE` works pretty much like `CLKOUT<n>_PHASE`, but on the input clock frequency and negative.

So I think the equations for the output clock wave forms are:
```
output clock period = CLKFBOUT_MULT / DIVCLK _DIVIDE / CLKOUT<n>_DIVIDE * input clock period
output clock rising edge = (input clock rising edge - (CLKFBOUT_PHASE / 360) * input clock period + (CLKOUT<n>_PHASE / 360) * output clock period) % output clock period
output clock falling edge = (output clock rising edge + CLKOUT<n>_DUTY_CYCLE * output clock period) % output clock period
```

litghost (commented):
Proposed fix for PLL phase shifts

tmichalak:
@litghost All of the above makes sense. Thanks a lot for the analysis. Judging from it the only things missing in the original implementation were the `period modulo` and negative phase shift from the CLKFBOUT_PHASE. 
BTW in the equation for the period:
`output clock period = input clock period * (CLKFBOUT_MULT / DIVCLK_DIVIDE) / CLKOUT<n>_DIVIDE `
it should be more like:
`output clock period = input clock period * (DIVCLK_DIVIDE * CLKOUT<n>_DIVIDE) / CLKFBOUT_MULT`
otherwise bigger frequency multiplier values would result in bigger periods

litghost:
> BTW in the equation for the period:
> `output clock period = input clock period * (CLKFBOUT_MULT / DIVCLK_DIVIDE) / CLKOUT<n>_DIVIDE `
> it should be more like:
> `output clock period = input clock period * (DIVCLK_DIVIDE * CLKOUT<n>_DIVIDE) / CLKFBOUT_MULT`
> otherwise bigger frequency multiplier values would result in bigger periods

Of course.  Higher period -> more division.

sdc-plugin/buffers.h
```diff
@@ -51,13 +51,8 @@ struct Pll {
 	clk_mult = FetchParam(cell, "CLKFBOUT_MULT", 5.0);
 	divclk_divisor = FetchParam(cell, "DIVCLK_DIVIDE", 1.0);
 	for (auto clk_output : outputs) {
-	    // CLKOUT[0-5]_DIVIDE
-	    clkout_divisors[clk_output] = FetchParam(cell, clk_output + "_DIVIDE", 1.0);
 	    clkout_period[clk_output] = CalculatePeriod(clk_output);
```
litghost, line 54: `CalculatePeriod` depends on "CLKIN1_PERIOD" being valid.  Should we add an explicit check that the clock period of "CLKIN1_PERIOD" matches the virtual clock constraint propigated to that net?

sdc-plugin/clocks.cc
```diff
@@ -226,8 +229,11 @@ void Clock::UpdatePeriod(float period) {
 void Clock::UpdateWaveform(float rising_edge, float falling_edge) {
     rising_edge_ = rising_edge;
     falling_edge_ = falling_edge;
-    if (falling_edge_ > period_) {
-	log_error("Phase shift on clock %s exceeds 360 degrees\nRising edge: %f, Falling edge: %f, Clock period:%f\n", name_.c_str(), rising_edge_, falling_edge_, period_);
+    if (rising_edge_ >= period_ ) {
```
litghost, line 232: If `ApplyShift` is called with a rising_edge greater than the period (which is legal), just subtracting 1 period isn't enough.

Also `ApplyShift` allows being invoked with a negative rising edge in the current code.  Maybe add an error if ApplyShift is called with a negative value?  If not, then add code to handle falling/rising edge < 0?

litghost:
Why is the delay through a Bufg still 1 ns?  We should either set it to 0, or a more reasonable value.  0 is probably better for now.

litghost:
Should have a FIXME/TODO about the `CLKFBOUT_PHASE` not being supported.  Maybe we should error if `CLKFBOUT_PHASE` is non-zero?

sdc-plugin/clocks.cc
tmichalak: Fixed by using `fmod` for period and waveform calculations.

sdc-plugin/buffers.h
tmichalak: Added the check.

tmichalak:
Added support for `CLKFBOUT_PHASE` and addressed all above mentioned comments regarding shift calculations.
The [period](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/37/files#diff-dae796c85b122bc4b7ed6a2aa41bd91dR59) and [waveform](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/37/files#diff-dae796c85b122bc4b7ed6a2aa41bd91dR68) is calculated according to the equations discussed in the discussion above.
Removed the temporary delay values on IBUF, BUFG (set them to 0).
Updated the references of all tests and added new ones for `DIVLK_DIVIDE` and `CLKFBOUT_PHASE`.

sdc-plugin/buffers.cc
```diff
@@ -15,10 +15,75 @@
...
 const std::vector<std::string> Pll::outputs = {"CLKOUT0", "CLKOUT1", "CLKOUT2",
                                                "CLKOUT3", "CLKOUT4", "CLKOUT5"};
 const float Pll::delay = 0;
 const std::string Pll::name = "PLLE2_ADV";
+
+Pll::Pll(RTLIL::Cell* cell, float input_clock_period, float input_clock_shift) {
+    assert(RTLIL::unescape_id(cell->type) == "PLLE2_ADV");
+
+    FetchParams(cell);
+    if (clkin1_period != input_clock_period) {
```
litghost, line 32: This should be a approx equal check, e.g. `fabs(a - b) < max(a, b)*10*epsilon`.  This prevents non-relevant rounding errors from triggering the error.

sdc-plugin/buffers.cc
```diff
@@ -15,10 +15,75 @@
...
+	// CLKOUT[0-5]_PHASE
+	clkout_phase[output] = FetchParam(cell, output + "_PHASE", 0.0);
+    }
+}
+
+void Pll::CalculateOutputClockPeriods() {
+    for (auto output : outputs) {
+	// CLKOUT[0-5]_PERIOD = CLKIN1_PERIOD * CLKOUT[0-5]_DIVIDE * DIVCLK_DIVIDE /
+	// CLKFBOUT_MULT
+	clkout_period[output] = clkin1_period * clkout_divisor.at(output) / clk_mult *
```
litghost, line 63: I think we should convert all uses of `clkin1_period` to be `float Pll::ClkinPeriod()`, and have it return `clkin1_period`, with a TODO on the `CLKINSEL`.

litghost: It's a little weird that `CalculateOutputClockPeriods` uses `clkin1_period` and `CalculateOutputClockWaveforms` uses `input_clock_period`.  I believe we should be consistent?

mithro:
For future developers I feel like this could benefit from a write up with a few diagrams (maybe wavedrom based)?

sdc-plugin/buffers.cc
tmichalak: Done. However I am not sure what is meant by the `10` in  `10*epsilon`.
I implemented it as `fabs(a-b) < max(a, b) * kApproxEqualFactor` where the `kApproxEqualFactor` is expressed in float, i.e. 0.01 means +/- 1%

sdc-plugin/buffers.cc
tmichalak: Done

tmichalak:
> For future developers I feel like this could benefit from a write up with a few diagrams (maybe wavedrom based)?

Wavedrom looks awesome!
I think at some point we will have to document what we have done in all of the Yosys plugins.
Wavedrom for explaining the SDC plugin will be indispensable.

sdc-plugin/buffers.cc
litghost: https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon

sdc-plugin/buffers.cc
```diff
@@ -15,10 +15,80 @@
...
+#include <cmath>
 #include "buffers.h"
 
 const std::vector<std::string> Pll::inputs = {"CLKIN1", "CLKIN2"};
 const std::vector<std::string> Pll::outputs = {"CLKOUT0", "CLKOUT1", "CLKOUT2",
                                                "CLKOUT3", "CLKOUT4", "CLKOUT5"};
 const float Pll::delay = 0;
 const std::string Pll::name = "PLLE2_ADV";
+
+Pll::Pll(RTLIL::Cell* cell, float input_clock_period, float input_clock_shift) {
```
litghost, line 28: `input_clock_shift` is poorly named?  It should be `input_clock_rising_edge`, yes?

sdc-plugin/buffers.h
```diff
@@ -40,73 +38,49 @@ struct IBuf : Buffer {
...
 
-    // CLKOUT[0-5]_PERIOD = CLKIN1_PERIOD * CLKOUT[0-5]_DIVIDE * DIVCLK_DIVIDE /
-    // CLKFBOUT_MULT
-    // TODO Check the value on CLKINSEL
-    float CalculatePeriod(const std::string& output) {
-	return clkin1_period * clkout_divisors.at(output) / clk_mult *
-	       divclk_divisor;
-    }
+    // Calculate the rising and falling edges of the output clocks
+    void CalculateOutputClockWaveforms(float input_clock_shift);
```
litghost, line 58: `CalculateOutputClockWaveforms` / `CalculateOutputClockPeriods` are private methods IMO, because they have to invoked in a specified order.  When would an external caller ever invoke them?

sdc-plugin/buffers.cc
tmichalak: Yes, fixed.

sdc-plugin/buffers.h
tmichalak: Right, fixed.

sdc-plugin/buffers.cc
tmichalak: Thanks, fixed.

litghost (approved):
LGTM

PR #40, opened 2020-09-24
Cell counting plugin

mkurc-ant:
This is a plugin that allows to return to the TCL interpreter number of selected modules, cells or wires. It is used in synthesis script for QuickLogic FPGAs.

### DIFF
--- a/get_count-plugin/get_count.cc
+++ b/get_count-plugin/get_count.cc
@@ -1,0 +1,133 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
+ *  Copyright (C) 2020  The Symbiflow Authors
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "kernel/register.h"
+#include "kernel/rtlil.h"
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+void register_in_tcl_interpreter(const std::string& command) {
+    Tcl_Interp* interp = yosys_get_tcl_interp();
+    std::string tcl_script = stringf("proc %s args { return [yosys %s {*}$args] }", command.c_str(), command.c_str());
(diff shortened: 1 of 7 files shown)


get_count-plugin/get_count.cc
```diff
@@ -0,0 +1,137 @@
...
+#include "kernel/rtlil.h"
+
+#ifndef YS_OVERRIDE
+#define YS_OVERRIDE override
+#endif
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+void register_in_tcl_interpreter(const std::string& command) {
```
tmichalak, line 31: While working on other plugins I came to realize that this is not needed as this is called during `yosys -import`. Remoe the function and its use.

get_count-plugin/get_count.cc
```diff
@@ -0,0 +1,137 @@
...
+    enum class ObjectType {
+        NONE,
+        MODULE,
+        CELL,
+        WIRE
+    };
+
+    GetCount () :
+        Pass("get_count", "Returns count of various selected object types to the TCL interpreter") {
+            register_in_tcl_interpreter(pass_name);
```
tmichalak, line 48: remove

get_count-plugin/get_count.cc
```diff
@@ -0,0 +1,137 @@
...
+            break;
+        default:
+            log_assert(false);
+        }
+
+        // Return the value as string to the TCL interpreter
+        std::string value = std::to_string(count);
+
+        Tcl_Obj* tclStr = Tcl_NewStringObj(value.c_str(), value.size());
+        Tcl_ListObjAppendElement(tclInterp, tclList, tclStr);
```
tmichalak, line 131: I don't think you need a tcl list here:
```
Tcl_Obj* tclStr = Tcl_NewStringObj(value.c_str(), -1);
Tcl_SetObjResult(tclInterp, tclStr);
```
should be enough.

get_count-plugin/get_count.cc
```diff
@@ -0,0 +1,137 @@
...
+        }
+        else {
+            log_error("Object type not specified!\n");
+        }
+
+        extra_args(a_Args, 2, a_Design);
+
+        // Get the TCL interpreter
+        Tcl_Interp* tclInterp = yosys_get_tcl_interp();
+        Tcl_Obj*    tclList = Tcl_NewListObj(0, NULL);
```
tmichalak, line 98: I don't think you need a tcl list, look comment below.

get_count-plugin/tests/simple/script.tcl
```diff
@@ -0,0 +1,30 @@
+yosys plugin -i ../../get_count.so
```
tmichalak, line 1: the path shouldn't require any `../..`

tmichalak (changes requested):
Looks good in general. Several comments in.

get_count-plugin/get_count.cc
mkurc-ant: Removed.

get_count-plugin/get_count.cc
mkurc-ant: Removed.

get_count-plugin/get_count.cc
mkurc-ant: Thanks, that works too. Fixed.

get_count-plugin/get_count.cc
mkurc-ant: Removed.

get_count-plugin/tests/simple/script.tcl
mkurc-ant: Fixed.

tmichalak: Oh, now I see. `plugin -i name.so` looks in the specified directory and falls back to Yosys's shared folder.
I don't think we want `install` to depend on `test`. The tests are run in CI after `install` which copies all .so to Yosys's shared folder thus `plugin -i get_count.so` will work in CI.  
With your Makefile conda_packages will run the test before installing in contrast to other plugins.
I don't think this is needed.
My suggestions:
1. Don't call test before install
2. Change `plugin -i ../../get_count.so` to `plugin -i get_count.so`. For completeness as both options will work in CI. However, what if your case requires another plugin? It won't be available before install.

mkurc-ant: Ok. Now I understand. Fixed.

tmichalak (approved):
LGTM, will merge on green CI.

PR #41, opened 2020-09-28
SDC: Add set_max_delay command

tmichalak:
This PR add basic support of the set_max_delay command.
It depends on #38 with the SdcWriter class.


litghost (commented):
CI is red.  Initial plugin seems ok, but I'm worried that we are missing something.  For now I believe let's let this through, and as problems arise we'll address them.

tmichalak:
> CI is red. Initial plugin seems ok, but I'm worried that we are missing something. 

CI is red because this PR depends on/shouldn't be merged before #38. 

> For now I believe let's let this through, and as problems arise we'll address them.

I fully agree with that

PR #42, opened 2020-09-28
Design introspection commands plugin

tmichalak:
This PR creates a plugin that contains the following commands:

- [x] get_nets
- [x] get_ports
- [x] get_pins
- [x] get_cells


litghost:
CI is red?

design_introspection-plugin/.clang-format
```diff
@@ -0,0 +1,151 @@
+---
```
litghost, line 1: Why is this being added in this subdirectory?

design_introspection-plugin/Makefile
```diff
@@ -0,0 +1,26 @@
+CXX = $(shell yosys-config --cxx)
+CXXFLAGS = $(shell yosys-config --cxxflags)
+LDFLAGS = $(shell yosys-config --ldflags)
```
litghost, line 3: Can we template this Makefile?  I keep seeing it over and over?

xdc-plugin/xdc.cc
```diff
@@ -345,10 +294,18 @@ struct SetProperty : public Pass {
 
 	// Extract signal name and port bit information from port name
 	std::pair<std::string, int> extract_signal(const std::string& port_name) {
+	        std::regex port_regex("\\{?([\\\\a-zA-Z0-9_\\.]+)(\\[([0-9]+)\\])?.*");
```
litghost, line 297: Does this match the underlying verilog/VHDL spec for ports?  This seems more restrictive than before

design_introspection-plugin/get_ports.h
```diff
@@ -0,0 +1,21 @@
...
+
+USING_YOSYS_NAMESPACE
+
+struct GetPorts : public GetCmd {
+    GetPorts() : GetCmd("get_ports", "Print matching ports") {}
+
+   private:
+    std::string TypeName() override;
+    std::string SelectionType() override;
+    /* void execute(std::vector<std::string> args, RTLIL::Design* design) override; */
```
litghost, line 14: Why is this commented line here?

design_introspection-plugin/get_ports.cc
```diff
@@ -0,0 +1,41 @@
...
+    int bit(0);
+    if (!sscanf(port_name.c_str(), "%[^[][%d]", &port_str[0], &bit)) {
+	log_error("Couldn't find port %s\n", port_name.c_str());
+    }
+
+    port_str.resize(strlen(port_str.c_str()));
+    RTLIL::IdString port_id(RTLIL::escape_id(port_str));
+    SelectionObjects objects;
+    if (auto wire = design->top_module()->wire(port_id)) {
+	if (wire->port_input || wire->port_output) {
```
litghost, line 26: What happens with `inout` ports?

design_introspection-plugin/get_ports.cc
```diff
@@ -0,0 +1,41 @@
...
+	    if (bit >= wire->start_offset &&
+	        bit < wire->start_offset + wire->width) {
+		objects.push_back(port_name);
+		if (!args.is_quiet) {
+		    log("%s ", port_name.c_str());
```
litghost, line 31: Not sure why verbosity should print when found, rather than not found?

design_introspection-plugin/get_cmd.h
```diff
@@ -0,0 +1,36 @@
...
+#define _GET_CMD_H_
+
+#include "kernel/register.h"
+
+USING_YOSYS_NAMESPACE
+
+struct GetCmd : public Pass {
+    using Filter = std::pair<std::string, std::string>;
+    using Filters = std::vector<Filter>;
+    using SelectionObjects = std::vector<std::string>;
```
litghost, line 11: Shouldn't you be using RTLIL string in this vector, because we are going to just send this to the TCL list, rather than a C++ object?

design_introspection-plugin/get_ports.cc
tmichalak: inout ports fall into both categories so they are treated in the same way as inputs or outputs

design_introspection-plugin/get_ports.cc
tmichalak: There are `-verbose` and `-quiet` switches in vivado commands. I don't think we need both. In my opinion `quiet` shouldn't print anything to the log while without `quiet` we should be as verbose as possible.

design_introspection-plugin/get_ports.h
tmichalak: Overlook, fixed.

design_introspection-plugin/Makefile
tmichalak: See PRs #43 and #44

design_introspection-plugin/.clang-format
tmichalak: I will finally work on the clang format PR to have it apply to all plugins.
I will remove it from this PR

tmichalak:
> CI is red?

Travis complained about exceeding the maximum log length.
It actually got to big to quickly find what failed.
I opened PRs #43 and #44

design_introspection-plugin/get_pins.cc
```diff
@@ -0,0 +1,60 @@
+#include "get_pins.h"
+
+USING_YOSYS_NAMESPACE
+
+std::string GetPins::TypeName() { return "pin"; }
+
+std::string GetPins::SelectionType() { return "c"; }
+
+void GetPins::ExecuteSelection([[gnu::unused]] RTLIL::Design* design,
```
mkurc-ant, line 9: Shouldn't it be more roboust to cast these unused args to void (like `(void)design;`) instead of using `[[gnu::unused]]` ?

design_introspection-plugin/get_cmd.cc
```diff
@@ -0,0 +1,128 @@
...
+	if (arg == "-filter" and argidx + 1 < args.size()) {
+	    std::string filter_arg = args[++argidx];
+
+	    // Remove spaces
+	    filter_arg.erase(
+	        std::remove_if(filter_arg.begin(), filter_arg.end(), isspace),
+	        filter_arg.end());
+
+	    // Parse filters
+	    std::regex filter_attr_regex("(\\w+\\s?==\\s?\\w+)([(||)(&&)]*)");
```
mkurc-ant, line 80: Please add here a comment describing what subset of the Vivado `-filter` syntax is supported. I can see that for now it is limited only to one `==` condition ?

design_introspection-plugin/get_cmd.cc
```diff
@@ -0,0 +1,128 @@
...
+	    std::regex_token_iterator<std::string::iterator> regex_end;
+	    std::regex_token_iterator<std::string::iterator> matches(
+	        filter_arg.begin(), filter_arg.end(), filter_attr_regex, 1);
+	    if (matches == regex_end) {
```
mkurc-ant, line 84: When this is true? You are comparing an initialized iterator with an uninitialized one ?

design_introspection-plugin/get_nets.cc
```diff
@@ -0,0 +1,30 @@
...
+
+std::string GetNets::TypeName() { return "net"; }
+
+std::string GetNets::SelectionType() { return "w"; }
+
+GetNets::SelectionObjects GetNets::ExtractSelection(RTLIL::Design* design,
+                                                    const CommandArgs& args) {
+    SelectionObjects selected_objects;
+    for (auto module : design->selected_modules()) {
+	for (auto wire : module->selected_wires()) {
```
mkurc-ant, line 13: Getting logical nets out of Yosys is more tricky, you need to examine cell connections. This command will return wires. A net may consist of multiple wires.

mkurc-ant (changes requested):

design_introspection-plugin/get_ports.cc
litghost: I think I disagree?  So to be clear, let's enumerate the possibilities from Vivado, please correct me:
 - Neither `-verbose` nor `-quiet`: Logging only occurs if the `get_X` fails to find anything, but filtering is used to limit output
 - `-quiet` is specified: Logging never occurs
 - `-verbose`: Logging only occurs if the `get_X` fails to find anything, no limiting is present

Under none of these does Vivado write retrieved results to the log?

tmichalak: Ok, for now I made sure that when `-quiet` is specified logging never occurs.

design_introspection-plugin/get_cmd.h
tmichalak: I don't really see a big benefit from using `RTLIL::IdString` in the current implementation. I would rather keep the dependency on the specific Yosys structs as low as possible.

design_introspection-plugin/get_pins.cc
tmichalak: Not sure, which one is more robust, but `(void)design` is more versatile so I will take the suggestion.

design_introspection-plugin/get_cmd.cc
tmichalak: It's true when no match is found since the default constructor constructs an end-of-sequence iterator. 
Refer to: http://www.cplusplus.com/reference/regex/regex_token_iterator/regex_token_iterator/

design_introspection-plugin/get_cmd.cc
tmichalak: Sure, added. Note that we even display a warning with this limitation.

design_introspection-plugin/get_nets.cc
tmichalak: This needs more investigation, let me open a separate issue for this, so other introspection commands are not blocked by this.

xdc-plugin/xdc.cc
tmichalak: This was a leftover from an experimental implementation. Removed.

mkurc-ant (approved):
LGTM

design_introspection-plugin/get_cmd.cc
```diff
@@ -0,0 +1,138 @@
...
+void GetCmd::help() {
+    //   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+    log("\n");
+    log("   get_%ss [-quiet] [-filter filter_expression] "
+        "<%s_selection> \n",
+        TypeName().c_str(), TypeName().c_str());
+    log("\n");
+    log("Get matching %ss\n", TypeName().c_str());
+    log("\n");
+    log("Print the output to stdout too. This is useful when all Yosys "
```
litghost, line 14: This part of the documentation is out of date?

litghost (approved):
LGTM

PR #43, opened 2020-10-06
Create common Makefile template to be used by all plugins

tmichalak:
Makefiles in the plugins have a lot in common. 
This PR creates a common Makefile and modifies all plugin's Makefiles to use the common one.


mkurc-ant (approved):
LGTM

Makefile_plugin.common
```diff
@@ -0,0 +1,26 @@
+CXX = $(shell yosys-config --cxx)
+CXXFLAGS = $(shell yosys-config --cxxflags)
+LDFLAGS = $(shell yosys-config --ldflags)
+LDLIBS = $(shell yosys-config --ldlibs)
+PLUGINS_DIR = $(shell yosys-config --datdir)/plugins
+
+SOURCES := $(shell find -name '*.cc')
+OBJS := $(SOURCES:cc=o)
+NAME := $(patsubst %-plugin,%,$(notdir $(shell pwd)))
```
litghost, line 9: Let's make `NAME` explicit, rather than implicit.

Makefile_plugin.common
```diff
@@ -0,0 +1,26 @@
+CXX = $(shell yosys-config --cxx)
+CXXFLAGS = $(shell yosys-config --cxxflags)
+LDFLAGS = $(shell yosys-config --ldflags)
+LDLIBS = $(shell yosys-config --ldlibs)
+PLUGINS_DIR = $(shell yosys-config --datdir)/plugins
+
+SOURCES := $(shell find -name '*.cc')
```
litghost, line 7: Please make `SOURCES` explicit.

tmichalak: Done.

Makefile_plugin.common
tmichalak: Done

litghost:
Needs a rebase/conflict resolved?

Makefile_plugin.common
```diff
@@ -0,0 +1,24 @@
+CXX = $(shell yosys-config --cxx)
```
litghost, line 1: Documentation please?

litghost: Should also mention how tests are invoked in the documentation?

xdc-plugin/Makefile
```diff
@@ -1,30 +1,10 @@
...
-	$(MAKE) -C tests all
-
-.PHONY: install
-install: install_modules install_plugin
-
-clean:
-	rm -f *.d *.o xdc.so
-	$(MAKE) -C tests clean
-
+install: install_modules
```
litghost, line 10: I believe this will make `make` made about the double definition of `install`.  Should use `install::` instead of `install:` in all instances?  As an alternative, you could create a variable like "EXTRA_INSTALL_TARGETS" or something.

tmichalak: I was also worried about that, but as it turned out without a reason. No warning and it does what it is supposed to

litghost: Ok, cool, lets leave as is.

litghost (changes requested):
Makefile template needs documentation still.

tmichalak:
> Makefile template needs documentation still.

Added some documentation. Is there anything else we want to describe in there?

Makefile_plugin.common
litghost: This is not the documentation I was expecting.  I was looking for:
 - What this template is for?
 - How to use it?
 - How to test with it?

tmichalak: @litghost How about now?

litghost: I want a description of how a developer can add automated test cases, e.g. make a directory called tests, consult this other makefile template, etc.

I think the idea here is that if a new developer comes in, how do they add a plugin without reading all the makefile itself.

I think showing an example directory structure for the makefile example plus referencing the testing makefile template will finish this off.

tmichalak: Mentioned the Makefile_test.common and added an example directory structure.

PR #44, opened 2020-10-06
Create common test Makefile template

tmichalak:
Makefiles in the test directories of the plugins have a lot in common.
This PR creates a common test Makefile and modifies all plugin's Makefiles to use the common one.
Moreover it executes the test with `-quiet` switch as Travis started to fail due to exceeding log length.


Makefile_test.common
```diff
@@ -0,0 +1,37 @@
+define test_tpl =
```
mkurc-ant, line 1: I would just rename it and call it a "common test" but rather a "common json diff test". Some plugin may use different testing strategies.

Makefile_test.common
```diff
@@ -0,0 +1,37 @@
...
+	RETVAL=$$$$? ; \
+	if [ $$$$RETVAL -eq 0 ]; then \
+		echo "Test $(1) PASSED"; \
+		touch $$<; \
+		true; \
+	else \
+		echo "Test $(1) FAILED"; \
+		false; \
+	fi
+
```
litghost, line 13: This needs sufficient documentation so that future developers understand how to use this.

Makefile_test.common
```diff
@@ -0,0 +1,37 @@
...
+endef
+compare = diff $(1)/$(1).golden.$(2) $(1)/$(1).$(2)
+compare_json = python compare_output_json.py --json $(1)/$(1).json --golden $(1)/$(1).golden.json
+define json_tpl =
+$(1)_update_json:
+	python compare_output_json.py --json $(1)/$(1).json --golden $(1)/$(1).golden.json --update
+endef
+
+TESTS = $(subst /, ,$(wildcard */))
```
litghost, line 28: Please make `TESTS` explicit as an input, rather than implicit as a glob/wildcard

Makefile_test.common
tmichalak: Removed

Makefile_test.common
tmichalak: Added

Makefile_test.common
tmichalak: Done

fasm-plugin/tests/Makefile
```diff
@@ -1,6 +1 @@
-TESTS = dummy
-
-all: $(TESTS)
-
-dummy:
-	@echo $@ PASS
+include $(shell pwd)/../../Makefile_test.common
```
litghost, line 1: This needs a test?  Or it just doesn't have a test?

tmichalak: It does not have a test. We should open an issue to make one. From what I recall that was the first plugin and we somehow didn't add a testsuite.

litghost: Please create an issue, but I didn't believe this required blocking the merge, etc.

Issue #45, opened 2020-10-08
Some plugins hold live pointers beyond liveness scope


litghost:
The current implementation of the SDC plugin holds RTLIL objects between TCL commands.  Because of this, changes to those pointers (e.g. techmap passes, etc) cause those pointers to become invalid.

Test case for the SDC plugin attached: [test.zip](https://github.com/SymbiFlow/yosys-symbiflow-plugins/files/5351665/test.zip)

litghost:
There may be more examples of this.  In general, passes can only ensure liveness of RTLIL objects within the scope of the pass, and no longer.  Holding these pointers between invocations is not safe, and will result in memory corruption.

litghost:
@tmichalak This is the highest priority issue, and must be resolved.  This is not optional.

tmichalak:
@litghost so for SDC the issue is that we are storing pointers to RTLIL::Wire* in the Clock objects, right?
In case Yosys destroys the wires what should be the desired behavior apart from not crashing (in this a simple !=NULL will do)?
Removing the Wire and clock if that is the last or only one clock wire?

litghost:
> @litghost so for SDC the issue is that we are storing pointers to RTLIL::Wire* in the Clock objects, right?
> In case Yosys destroys the wires what should be the desired behavior apart from not crashing (in this a simple !=NULL will do)?
> Removing the Wire and clock if that is the last or only one clock wire?

You cannot hold the pointers at all.  It is not a safe behavior.  My recommendation is to defer module inspection until when you are ready to do so (e.g. write_sdc).  Having intermediate `RTLIL::Wire*` objects between TCL calls/yosys passes will always be unsafe.

Pointer checks are useless here, because yosys deleting an address won't update the pointer copy held in the Clocks object.

tmichalak:
@litghost I created a draft PR targeting the issue.
Basically, I created a write_sdc script pass command which wraps the `propagate_clocks` and `write_sdc` commands effectively postponing the clock propagation to when `write_sdc` is called.
As for the pointers, I am just cleaning the clock structure after each run, so this is not optimal since every `write_sdc` triggers a new clock propagation step.
I am aware that we can't keep pointers between runs, so maybe instead of keeping `RTLIL::Wire*` objects, I could just keep the wire name. 
However, in the test you added to the issue I think that cleaning the clock structure is actually the desired behavior.
[...]

PR #49, opened 2020-10-19
Write sdc multiple pass

tmichalak:
This PR present a possible fix for the bug reported in issue #45.
It creates a `write_sdc` script pass command which wraps the `propagate_clocks` and write_sdc` commands effectively postponing the clock propagation to happen when `write_sdc` is called, right before actaully writing out the final SDC file.

### DIFF
--- a/sdc-plugin/clocks.h
+++ b/sdc-plugin/clocks.h
@@ -73,6 +73,9 @@ class Clocks {
     const std::vector<Clock> GetClocks() {
 	return clocks_;
     }
+    void Clear() {
+	clocks_.clear();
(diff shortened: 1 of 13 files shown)


litghost (changes requested):
This doesn't solve the underlying issue, which is that `RTLIL` objects are being held between script commands.  For example, `get_clocks` stores pointers, which `write_sdc` then use later.  Those pointers do not have to remain valid between script invocations, and in some cases are invalidated.

Did you test the `valgrind` case I provided in #45 and show that it now passes?

PR #50, opened 2020-10-19
Add -through switch to set_false_path command

tmichalak:
The XDC file generated for the base litex design contains the following command:
```
set_false_path -quiet -through [get_nets -hierarchical -filter {mr_ff == TRUE}]
```
While we do have the `set_false_path` command already implemented we are missing the `through` switch.
This PR addressed that.


litghost (approved):
LGTM

PR #51, opened 2020-10-23
Keep the SDC information in Yosys's design structures

### LINKED ISSUE #45: Some plugins hold live pointers beyond liveness scope

tmichalak:
This PR fixes #45 by modifying the current SDC plugin implementation to not keep the clock and generated clock information in a separate structure, but in Yosys' internal design representation. For this purpose the clock related attributes from [this list](https://docs.google.com/spreadsheets/d/1G-E2Dq8YG4g9Z6mTygpumwlI_vNlFUQinc9gMgePfec/edit#gid=1349094793) on the specified wires are set or updated using `create_clocks` or as a result of clock propagation.

### DIFF
--- a/Makefile_test.common
+++ b/Makefile_test.common
@@ -12,6 +12,12 @@
 # test1_verify = $(call diff_test,test1,ext) && test $$(grep "PASS" test1/test1.txt | wc -l) -eq 2
 # test2_verify = $(call diff_test,test2,ext)
 #
+GTEST_DIR = ../../third_party/googletest/googletest
+CXX = $(shell yosys-config --cxx)
+CXXFLAGS = $(shell yosys-config --cxxflags) -I.. -I$(GTEST_DIR)/include
+LDLIBS = $(shell yosys-config --ldlibs) -L$(GTEST_DIR)/build/lib -lgtest -lgtest_main -lpthread
+LDFLAGS = $(shell yosys-config --ldflags)
+
 define test_tpl =
 $(1): $(1)/ok
 	@echo "Verifying result of test $(1)"
@@ -50,13 +56,34 @@ $(1)/ok: $(1)/$(1).v
 
 endef
 
+define unit_test_tpl =
+$(1): $(1)/$(1).test
+	@$$<
+
+$(1)/$(1).test: $(1)/$(1).test.o $$(GTEST_DIR)/build/lib/libgtest.a
+	@$(CXX) $(LDFLAGS) -o $$@ $$< $(LDLIBS)
+
+$(1)/$(1).test.o: $(1)/$(1).test.cc
+	@$(CXX) $(CXXFLAGS) $(LDFLAGS) -c $$< -o $$@
+
+endef
+
 diff_test = diff $(1)/$(1).golden.$(2) $(1)/$(1).$(2)
 
-all: $(TESTS)
-.PHONY: all clean $(TESTS)
+all: $(TESTS) $(UNIT_TESTS)
+
+$(GTEST_DIR)/build/lib/libgtest.a $(GTEST_DIR)/build/lib/libgtest_main.a:
+	@mkdir -p $(GTEST_DIR)/build
+	@cd $(GTEST_DIR)/build; \
+	cmake ..; \
+	make
+
+.PHONY: all clean $(TESTS) $(UNIT_TESTS)
 
 $(foreach test,$(TESTS),$(eval $(call test_tpl,$(test))))
+$(foreach test,$(UNIT_TESTS),$(eval $(call unit_test_tpl,$(test))))
 
 clean:
 	@rm -rf $(foreach test,$(TESTS),$(test)/$(test).sdc $(test)/$(test).txt $(test)/$(test).eblif $(test)/$(test).json)
+	@rm -rf $(foreach test,$(UNIT_TESTS),$(test)/$(test).test.o $(test)/$(test).test.d $(test)/$(test).test)
 	@find . -name "ok" -or -name "*.log" | xargs rm -rf
--- a/Makefile_test.common
+++ b/Makefile_test.common
@@ -33,7 +33,7 @@ $(1)/ok: $(1)/$(1).v
 
 endef
 
-diff_test = bash -c "diff <(sort $(1)/$(1).golden.$(2)) <(sort $(1)/$(1).$(2))"
+diff_test = diff $(1)/$(1).golden.$(2) $(1)/$(1).$(2)
 
 all: $(TESTS)
 .PHONY: all clean $(TESTS)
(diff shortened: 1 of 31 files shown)


Makefile_test.common
```diff
@@ -33,7 +33,7 @@ $(1)/ok: $(1)/$(1).v
 
 endef
 
-diff_test = diff $(1)/$(1).golden.$(2) $(1)/$(1).$(2)
```
litghost, line 36: The diff test should not sort.  Normalization should occur on the plugin side, so that for the same inputs, you get the same outputs. @mithro , thoughts?

mithro: I concur that it would be better for the output sdc file to be produced in sorted order if possible.

tmichalak: Added sorting to SDC output and reverted sorting in diff_test.

sdc-plugin/clocks.cc
```diff
@@ -19,220 +19,104 @@
...
+	    "used\n",
+	    falling_edge);
+	return std::make_pair(0, falling_edge);
     }
-    return res;
+    float rising_edge(0);
+    float falling_edge(0);
+    std::string waveform(
+        clock_wire->get_string_attribute(RTLIL::escape_id("WAVEFORM")));
+    std::sscanf(waveform.c_str(), "%f %f", &rising_edge, &falling_edge);
```
litghost, line 81: You are missing error handing here?

sdc-plugin/clocks.cc
```diff
@@ -19,220 +19,104 @@
...
-	std::stringstream ss;
-	for (auto clock_wire : clock.GetClockWires()) {
-	    ss << RTLIL::unescape_id(clock_wire->name) << " ";
+float Clock::Period(RTLIL::Wire* clock_wire) {
+    if (!clock_wire->has_attribute(RTLIL::escape_id("PERIOD"))) {
+	log_warning(
+	    "Period has not been specified\n Default value 0 will be used\n");
+	return 0;
+    }
+    return std::stof(
```
litghost, line 57: If this throws an exception, does yosys handle it well?

sdc-plugin/clocks.cc
```diff
@@ -19,220 +19,104 @@
...
-	    AddClock(RTLIL::unescape_id(wire->name), wire, clock.Period(),
-	             clock.RisingEdge() + path_delay,
-	             clock.FallingEdge() + path_delay);
-	}
+std::string Clock::WireName(RTLIL::Wire* wire) {
+    if (!wire) {
+	return std::string();
     }
+    std::string wire_name(RTLIL::unescape_id(wire->name));
+    return std::regex_replace(wire_name, std::regex{"\\$"}, "\\$");
```
litghost, line 105: Can you explain what this regex is doing?  Because at first blush, this doesn't do anything?

litghost (commented):
Comments in

sdc-plugin/clocks.cc
tmichalak: I will add an comment, but in brief it makes sure we escape `$` in the output sdc e.g. we get
`\$auto\$clkbufmap.cc:247:execute\$1829`
instead of
`$auto$clkbufmap.cc:247:execute$1829`

sdc-plugin/clocks.cc
tmichalak: Added along with negative test

sdc-plugin/clocks.cc
tmichalak: No, it didn't. Added code to catch the incorrect argument exception along with a negative test.

sdc-plugin/clocks.cc
```diff
@@ -19,220 +19,111 @@
...
-	    AddClock(RTLIL::unescape_id(wire->name), wire, clock.Period(),
-	             clock.RisingEdge() + path_delay,
-	             clock.FallingEdge() + path_delay);
-	}
+    float rising_edge(0);
+    float falling_edge(0);
+    std::string waveform(
+        clock_wire->get_string_attribute(RTLIL::escape_id("WAVEFORM")));
+    if (std::sscanf(waveform.c_str(), "%f %f", &rising_edge, &falling_edge) != 2) {
+	log_cmd_error("Incorrect WAVEFORM format\n");
```
litghost, line 87: Please print the invalid WAVEFORM too?

litghost: More specifically, error messages should:
 - Say what is wrong
 - Include enough context to determine why the input was wrong
 - Provide enough context to determine what is right

In this case:
 - The WAVEFORM attribute was wrong (good)
 - However which wire had the bad attribute is not given (bad)
 - The bad attribute is not given (bad)
 - What a good WAVEFORM looks like is not given (bad)

Please improve the error message.

sdc-plugin/clocks.cc
litghost: Would it be possible to factor this into a small function, and you write some unit tests for that small function?  Thanks

litghost (commented):
Comments in

sdc-plugin/clocks.cc
tmichalak: Added googletest and some Makefile template to perform unit tests.

sdc-plugin/clocks.cc
tmichalak: Done.

sdc-plugin/clocks.cc
```diff
@@ -19,220 +19,121 @@
...
+    if (!clock_wire->has_attribute(RTLIL::escape_id("WAVEFORM"))) {
+	float period(Period(clock_wire));
+	if (!period) {
+	    log_cmd_error(
+	        "Neither PERIOD nor WAVEFORM has been specified for wire %s\n",
+	        WireName(clock_wire).c_str());
+	    return std::make_pair(0, 0);
 	}
+	float falling_edge = period / 2;
+	log_warning(
```
litghost, line 82: This warning and the warning on line 53 probably should indicate which wire they are referring too.

sdc-plugin/clocks.cc
litghost: Thanks

sdc-plugin/clocks.cc
tmichalak: Fixed

litghost (approved):
LGTM

litghost:
@mithro Do you want to review this?

mithro:
Nope, proceed without me!

PR #52, opened 2020-10-27
Add -of switch and -include_generated_clocks to get_clocks command

tmichalak:
This PR depends on #51 
It add the `-of` and `-include_generated_clocks` switches to `get_clocks` command since they are used in the linux litex SDC.

### DIFF
--- a/sdc-plugin/sdc.cc
+++ b/sdc-plugin/sdc.cc
@@ -21,10 +21,10 @@
 #include "kernel/register.h"
 #include "kernel/rtlil.h"
 #include "propagation.h"
+#include "sdc_writer.h"
+#include "set_clock_groups.h"
 #include "set_false_path.h"
 #include "set_max_delay.h"
-#include "set_clock_groups.h"
-#include "sdc_writer.h"
 
 USING_YOSYS_NAMESPACE
 
@@ -73,7 +73,8 @@ struct WriteSdcCmd : public Backend {
     }
 
     void execute(std::ostream*& f, std::string filename,
-                 std::vector<std::string> args, RTLIL::Design* design) override {
+                 std::vector<std::string> args,
+                 RTLIL::Design* design) override {
 	if (args.size() < 2) {
 	    log_cmd_error("Missing output file.\n");
 	}
@@ -86,8 +87,7 @@ struct WriteSdcCmd : public Backend {
 };
 
 struct CreateClockCmd : public Pass {
-    CreateClockCmd()
-        : Pass("create_clock", "Create clock object") {}
+    CreateClockCmd() : Pass("create_clock", "Create clock object") {}
 
     void help() override {
 	log("\n");
@@ -169,8 +169,7 @@ struct CreateClockCmd : public Pass {
 	    rising_edge = 0;
 	    falling_edge = period / 2;
 	}
-	Clock::Add(name, selected_wires, period, rising_edge,
-	                 falling_edge);
+	Clock::Add(name, selected_wires, period, rising_edge, falling_edge);
     }
 
     void AddWirePrefix(std::vector<std::string>& args, size_t argidx) {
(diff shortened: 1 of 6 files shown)


sdc-plugin/tests/get_clocks/get_clocks.v
```diff
@@ -0,0 +1,36 @@
+module top(input clk,
+        input clk2,
+	input [1:0] in,
+	output [5:0] out );
+
+reg [1:0] cnt = 0;
+wire clk_int_1, clk_int_2;
+IBUF ibuf_proxy(.I(clk), .O(ibuf_proxy_out));
+IBUF ibuf_inst(.I(ibuf_proxy_out), .O(ibuf_out));
```
litghost, line 9: This structure isn't legal?  What are you trying to test?

sdc-plugin/sdc.cc
```diff
@@ -192,18 +192,59 @@ struct GetClocksCmd : public Pass {
...
+
+	    break;
 	}
+	std::vector<std::string> clocks_list(args.begin() + argidx, args.end());
 	std::map<std::string, RTLIL::Wire*> clocks(Clocks::GetClocks(design));
 	if (clocks.size() == 0) {
 	    log_warning("No clocks found in design\n");
 	}
+	for (auto c : clocks_nets) {
+	    log("Clock filter %s\n", c.c_str());
```
litghost, line 231: This should be behind a verbose flag or some such?

sdc-plugin/sdc.cc
```diff
@@ -192,18 +192,59 @@ struct GetClocksCmd : public Pass {
 	log("\n");
     }
 
+    std::vector<std::string> extract_list(const std::string& args) {
```
litghost, line 195: Is there a way to leverage the TCL interpreter's list processing here?  This doesn't seem bad, but I wonder if there will be some TCL <-> this implemenation mismatch?

sdc-plugin/sdc.cc
```diff
@@ -192,18 +192,59 @@ struct GetClocksCmd : public Pass {
...
+	std::istream_iterator<std::string> end;
+	std::copy(begin, end, std::back_inserter(port_list));
+	return port_list;
+    }
+
     void execute(std::vector<std::string> args,
                  RTLIL::Design* design) override {
-	if (args.size() > 1) {
-	    log_warning("Command doesn't support arguments, so they will be ignored.\n");
+	bool generated_clocks(false);
```
litghost, line 206: Can you please add some more comments through here?

sdc-plugin/sdc.cc
tmichalak: We probably could, but I am filtering the content of the list later on before packing the final TCL list, so the use of a Tcl_Object would be a bit artificial in my opinion.

sdc-plugin/sdc.cc
tmichalak: Add a debug macro I am using in other SDC source files.

sdc-plugin/sdc.cc
tmichalak: You mean the generated_clocks part. Added a comment we currently treat all added clocks as auto-generated.

sdc-plugin/tests/get_clocks/get_clocks.v
tmichalak: You are right, this was an artificial structure. Fixed.

sdc-plugin/sdc.cc
litghost: I guess my point is that if the `list` that this function is extracting is a Tcl list, I suspect believe your implementation does not respects the Tcl parsing rules properly.

litghost:
CI is red?

sdc-plugin/tests/get_clocks/get_clocks.golden.txt
```diff
@@ -0,0 +1,4 @@
+
```
litghost, line 1: This first line is definitely wrong?  `get_clocks` should return something???

sdc-plugin/tests/get_clocks/get_clocks.tcl
```diff
@@ -0,0 +1,31 @@
...
+# Write the clocks to file
+set fh [open $::env(DESIGN_TOP).txt w]
+
+puts $fh [get_clocks]
+
+puts $fh [get_clocks -include_generated_clocks]
+
+puts $fh [get_clocks -include_generated_clocks clk2]
+
+puts $fh [get_clocks -of [get_nets clk_int_1 clk1] -include_generated_clocks clk_int_1]
```
litghost, line 29: Should have a test that is just `-of [get_nets ...]`, yes?

sdc-plugin/sdc.cc
litghost: Am I missing something, or did you not add a comment?

litghost (changes requested):
Comments in

sdc-plugin/sdc.cc
tmichalak: I don't know what type of comment should go there. The switch itself is described in the tool's [help](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/52/files#diff-e71c67a80542c63bb4832fffed080e0d86d1cc0dee1ac1b0e255fd7f6520e0aeR192) and some comment about generated clocks [here](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/52/files#diff-e71c67a80542c63bb4832fffed080e0d86d1cc0dee1ac1b0e255fd7f6520e0aeR213)

tmichalak:
> CI is red?

Fixed

sdc-plugin/tests/get_clocks/get_clocks.tcl
tmichalak: [Added](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/52/files#diff-66206c3e9c4388d75399de4a47827146bf328f9738dd60b32606c948c9175d86R31)

sdc-plugin/tests/get_clocks/get_clocks.golden.txt
tmichalak: Fixed.

sdc-plugin/sdc.cc
tmichalak: Could you give an example? `get_clocks -of [list a b c]` and `get_clocks -of [get_nets ...]` works in the current implementation.

sdc-plugin/sdc.cc
litghost: Your code below this comment is doing a bunch of stuff.  What is it doing, and why it is doing it?

sdc-plugin/sdc.cc
litghost: `get_clocks -of [concat [get_nets c] [get_nets {a b}]]` ?

tmichalak: > `get_clocks -of [concat [get_nets c] [get_nets {a b}]] `

[this](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/52/files#diff-66206c3e9c4388d75399de4a47827146bf328f9738dd60b32606c948c9175d86R33) works

sdc-plugin/tests/get_clocks/get_clocks.golden.txt
```diff
@@ -0,0 +1,6 @@
+{$auto$clkbufmap.cc:247:execute$1917} {$auto$clkbufmap.cc:247:execute$1919} clk clk2 clk_int_1 middle_inst_1.clk_int middle_inst_4.clk
```
litghost, line 1: There doesn't appear to be a difference between line 1 and 2?  What is the expected behavior of `-include_generated_clocks`, because I don't see you actually testing that?

litghost: I think we may have a conceptual interface mismatch with how Vivado treats clock objects.  The clock object on both sides of a IBUF and BUFG are the same clock object.  So in this example, there should really just be 1 clock `clk`, which has the nets listed on line 1 as the nets that are on that clock.  PLL's / MMCM's and BUFGCTRL's do split clock definitions because the underlying properties of the clock change across the elements.  The propigation that occurs should probably be something akin too:
 - Does this element change the clock -> No:
   - If the output net has no explicit clock membership, mark the net as being part of the same clock
 - Does this element change the clock -> Yes:
  - If the output net has an explicit clock definition, do nothing
  - If the output net has no explicit clock definition, create a new generated clock with the new properties based on the elements configuration (e.g. PLL creates a new clock with period X, waveform Y Z).

Does that make sense?

litghost: To phrase it a different way:
 - Clock objects has properties (e.g. waveform, period, is generated)
 - Clocks can have relationships to each other (e.g. false path, jitter, https://docs.verilogtorouting.org/en/latest/vpr/sdc_commands/#set-max-delay-set-min-delay)
 - Nets may have a clock object they are associated with.  The propigation step should associate nets with their clocks as needed (e.g. propagate through IBUF/BUFG).
 - Some cells (PLL/MMCM) define generated clocks using a relationship between the source clock and output clock.  The propagate through PLL/MMCM should generate these related clocks.

sdc-plugin/sdc.cc
tmichalak: Done

sdc-plugin/tests/get_clocks/get_clocks.golden.txt
tmichalak: > There doesn't appear to be a difference between line 1 and 2? What is the expected behavior of -include_generated_clocks, because I don't see you actually testing that?

Currently there is no difference between these two lines as all clocks are treated as non-generated, hence they are never skipped if the -include_generated_clocks is not specified.
Once issue #53 is implemented there will be differences, but for now we simply test the command interface.

tmichalak: @litghost I agree with the clock concept you described above. I opened issue #53 to capture what you described as I believe it's beyond the scope of this PR.

sdc-plugin/sdc.cc
```diff
@@ -181,29 +180,103 @@ struct CreateClockCmd : public Pass {
...
+    std::vector<std::string> extract_list(const std::string& args) {
+	std::vector<std::string> port_list;
+	std::stringstream ss(args);
+	std::istream_iterator<std::string> begin(ss);
+	std::istream_iterator<std::string> end;
+	std::copy(begin, end, std::back_inserter(port_list));
+	return port_list;
+    }
+
+    // TODO Check for GENERATED_CLOCK clock wire attribute
```
litghost, line 213: Add link to issue that will fix this issue.

sdc-plugin/tests/get_clocks/get_clocks.tcl
```diff
@@ -0,0 +1,35 @@
...
+plugin -i design_introspection
+# Import the commands from the plugins to the tcl interpreter
+yosys -import
+
+read_verilog $::env(DESIGN_TOP).v
+read_verilog -specify -lib -D_EXPLICIT_CARRY +/xilinx/cells_sim.v
+read_verilog -lib +/xilinx/cells_xtra.v
+hierarchy -check -auto-top
+# Start flow after library reading
+synth_xilinx -vpr -flatten -abc9 -nosrl -nodsp -iopad -run prepare:check
```
litghost, line 12: Drop the `-vpr` flag as we no longer use it.  Actually I recommend we consider testing with just `synth_xilinx` with no flags.

sdc-plugin/sdc.cc
tmichalak: Done

sdc-plugin/tests/get_clocks/get_clocks.tcl
tmichalak: Dropped the `-vpr` flag

Issue #53, opened 2020-11-03
SDC: Improve clock propigation and clock object handling


tmichalak:
I think we may have a conceptual mismatch with how Vivado treats clock objects. The clock object on both sides of a IBUF and BUFG are the same clock object. So for example, there should really just be 1 clock clk, which has multiple netss that are on that clock. PLL's / MMCM's and BUFGCTRL's do split clock definitions because the underlying properties of the clock change across the elements. 

Clock objects have properties (e.g. waveform, period, is generated)
Clocks can have relationships to each other (e.g. false path, jitter, https://docs.verilogtorouting.org/en/latest/vpr/sdc_commands/#set-max-delay-set-min-delay)
Nets may have a clock object they are associated with. The propagation step should associate nets with their clocks as needed (e.g. propagate through IBUF/BUFG).
Some cells (PLL/MMCM) define generated clocks using a relationship between the source clock and output clock. The propagate through PLL/MMCM should generate these related clocks.

The propagation that occurs should probably be something akin too:

- Does this element change the clock -> No:
  - If the output net has no explicit clock membership, mark the net as being part of the same clock

- Does this element change the clock -> Yes:
  - If the output net has an explicit clock definition, do nothing
  - If the output net has no explicit clock definition, create a new generated clock with the new properties based on the elements configuration (e.g. PLL creates a new clock with period X, waveform Y Z).

tmichalak:
Solved by #54

PR #54, opened 2020-11-23
Clock propagation enhancement

tmichalak:
This PR aims at solving issue #53.

Added a distinction between three types of clocks:
* explicit - added with `create_clock` command 
* generated - propagated clocks that have different clock parameters than the master clock (like PLLs or MMCMs)
* propagated clocks - clock wires with the same clock parameters as their master clock

### DIFF
--- a/sdc-plugin/clocks.h
+++ b/sdc-plugin/clocks.h
@@ -46,6 +46,7 @@ class Clock {
     static std::string AddEscaping(const std::string& name) {
 	return std::regex_replace(name, std::regex{"\\$"}, "\\$");
     }
+    static std::string SourcePinName(RTLIL::Wire* clock_wire);
 
    private:
     static std::pair<float, float> Waveform(RTLIL::Wire* clock_wire);
--- a/sdc-plugin/clocks.h
+++ b/sdc-plugin/clocks.h
@@ -32,12 +32,13 @@ class Propagation;
 
 class Clock {
    public:
+    enum ClockType { EXPLICIT, GENERATED, PROPAGATED };
     static void Add(const std::string& name, RTLIL::Wire* wire, float period,
-                    float rising_edge, float falling_edge);
+                    float rising_edge, float falling_edge, ClockType type);
     static void Add(const std::string& name, std::vector<RTLIL::Wire*> wires,
-                    float period, float rising_edge, float falling_edge);
+                    float period, float rising_edge, float falling_edge, ClockType type);
     static void Add(RTLIL::Wire* wire, float period, float rising_edge,
-                    float falling_edge);
+                    float falling_edge, ClockType type);
     static float Period(RTLIL::Wire* clock_wire);
     static float RisingEdge(RTLIL::Wire* clock_wire);
     static float FallingEdge(RTLIL::Wire* clock_wire);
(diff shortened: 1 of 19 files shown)


mithro:
I haven't actually looked at this pull request, but do please make sure we have a bunch of documentation around this.

mithro:
Also, if there is a conflict, which one wins?

sdc-plugin/clocks.h
```diff
@@ -32,12 +32,13 @@ class Propagation;
 
 class Clock {
    public:
+    enum ClockType { EXPLICIT, GENERATED, PROPAGATED };
```
litghost, line 35: Please add a comment on what these are?

sdc-plugin/sdc_writer.cc
```diff
@@ -46,15 +46,19 @@ void SdcWriter::WriteSdc(RTLIL::Design* design, std::ostream& file) {
...
 	    continue;
 	}
+	// Write out only GENERATED and EXPLICIT clocks
+	if (Clock::IsPropagated(clock_wire)) {
+	    continue;
+	}
 	file << "create_clock -period " << Clock::Period(clock_wire);
 	file << " -waveform {" << Clock::RisingEdge(clock_wire) << " "
 	     << Clock::FallingEdge(clock_wire) << "}";
-	file << " " << Clock::WireName(clock_wire);
```
litghost, line 57: Can you explain why this is called `SourcePinName`, rather than `SourceWireName`?  Does Yosys have a first class cell pin object?

sdc-plugin/tests/counter/counter.golden.txt
```diff
@@ -1 +1,2 @@
-{$auto$clkbufmap.cc:247:execute$1918} {$auto$clkbufmap.cc:247:execute$1920} clk clk2 clk_int_1 ibuf_proxy_out middle_inst_1.clk_int middle_inst_4.clk
+clk clk2 clk_int_1
```
litghost, line 1: This still shows the `get_clocks` and `get_clocks -include_generated_clocks` as the same?  Is there a golden example that shows `-include_generated_clocks` doing the right thing?

litghost (commented):
Looks pretty good.

litghost:
Conflicts?

sdc-plugin/clocks.h
tmichalak: Done

sdc-plugin/tests/counter/counter.golden.txt
tmichalak: I enhanced the `get_clocks` example which shows the difference.

sdc-plugin/sdc_writer.cc
tmichalak: `SOURCE_PIN` was the attribute in Vivado, but we do use Wires instead of pins. Hence, I changed the attribute and method name accordingly.

tmichalak:
> Conflicts?

Resolved

sdc-plugin/sdc_writer.cc
litghost: Can yosys return the wire attached to a cell pin? If so, it might make more sense to stick with `SOURCE_PIN`, and change this code to get the wire from the source pin?

PR #55, opened 2020-11-24
Add plugin to integrate inverters with invertible pins

tmichalak:
This plugins integrates inverters with invertible pins (pins with invertible attribute set).
So in case there is an inverter driving a pin with `invertible_pin` attribute set to name of the parameter we want to toggle, the inverter is being removed and the value of the attribute toggled. 
It can be used to absorb the inverter on a clock input (for example ISERDES's CLKB) and toggling the IS_CLKB_INVERTED attribute.


mithro:
I would call this "inverter absorption" but it seems that upstream Yosys calls this "merging". See the ["C.79 memory_dff – merge input/output DFFs into memories"](http://www.clifford.at/yosys/cmd_memory_dff.html) command for example.

The opposite of that command seems to be ["C.82 memory_nordff – extract read port FFs from memories"](http://www.clifford.at/yosys/cmd_memory_nordff.html).

It would have been better if that above example was something like `memory_merge_dff` and `memory_extract_dff`...

I would name this `pin_merge_inv` and make sure there is an `pin_extract_inv` which does the opposite.

integrateinv-plugin/integrateinv.cc
```diff
@@ -0,0 +1,343 @@
...
+                             "with 'invertible_pin' attribute set") {}
+
+  void help() override {
+    log("\n");
+    log("    integrateinv [selection]");
+    log("\n");
+    log("This pass integrates inverters into cells that have ports with the\n");
+    log("'invertible_pin' attribute set. The attribute should contain name\n");
+    log("of a parameter controlling the inversion. Whenever an inverter\n");
+    log("of a parameter controlling the inversion. Whenever an inverter\n");
```
mithro, line 76: Same line twice?

tmichalak: Fixed.

litghost (changes requested):
CI is red?

tmichalak:
> CI is red?

Rebased, it's green now

PR #57, opened 2020-11-25
SDC: Skip adding clocks on dangling wires

tmichalak:
This PR fixes the problem found in https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1753#issuecomment-733165240
If the wire on the clock divider output doesn't drive any cells it won't be added to the list of generated clocks.

### DIFF
--- a/sdc-plugin/propagation.cc
+++ b/sdc-plugin/propagation.cc
@@ -114,6 +114,22 @@ RTLIL::Cell* Propagation::FindSinkCellOnPort(RTLIL::Wire* wire,
     return sink_cell;
 }
 
+bool Propagation::WireHasSinkCell(RTLIL::Wire* wire) {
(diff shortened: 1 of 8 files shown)


mithro:
@tmichalak - Can the dangling wire just be removed via running the `opt_clean` command before the SDC write out command?

From the docs - http://www.clifford.at/yosys/cmd_opt_clean.html
> This pass identifies wires and cells that are unused and removes them. Other
> passes often remove cells but leave the wires in the design or reconnect the
> wires but leave the old cells in the design. This pass can be used to clean up
> after the passes that do the actual work.

acomodi (approved):
LGTM

mithro (commented):
Can you add an explicit test for this, otherwise LGTM.

PR #58, opened 2020-11-26
ci: migrate from Travis CI to GH actions

acomodi:


acomodi:
@tmichalak FYI. I am unsure why the job is not starting here, but I presume this is because there is no `.github/workflow` directory on master?

I will open a new PR that enables the github workflow, which will be initially empty, and then rebase this PR when the other one gets merged, otherwise we can trust the GH action CI running on the branch itself (https://github.com/antmicro/yosys-symbiflow-plugins/actions/runs/385189968).

An additional note is that I had to do two changes to the tests:
- remove the `-vpr` flag
- change some tests' golden files as some nets changed their name (maybe due to the previous point?)

tmichalak (approved):
LGTM

Issue #64, opened 2020-12-09
SDC plugin does not output the BUFG out clocks anymore


acomodi:
With the clock propagation enhancements done in https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/54, the resulting SDC does not take into account the output clock from BUFGs connected to the PLL outputs.

This causes VPR to not correctly constrain clock signals, resulting in higher run-time and an `nan` CPD, such as in the following example:

```
PLLE2_ADV #(
        .CLKFBOUT_MULT(4'd12),
        .CLKIN1_PERIOD(10.0),
        .CLKOUT0_DIVIDE(5'd20),
        .CLKOUT0_PHASE(1'd0),
        .CLKOUT1_DIVIDE(3'd5),
        .CLKOUT1_PHASE(1'd0),
        .CLKOUT2_DIVIDE(3'd5),
        .CLKOUT2_PHASE(7'd90),
        .CLKOUT3_DIVIDE(3'd6),
        .CLKOUT3_PHASE(1'd0),
        .CLKOUT4_DIVIDE(6'd48),
        .CLKOUT4_PHASE(1'd0),
        .DIVCLK_DIVIDE(1'd1),
        .REF_JITTER1(0.01),
        .STARTUP_WAIT("FALSE")
) PLLE2_ADV (
        .CLKFBIN(pll_fb),
        .CLKIN1(crg_clkin),
        .RST(reset7),
        .CLKFBOUT(pll_fb),
        .CLKOUT0(crg_clkout0),
        .CLKOUT1(crg_clkout1),
        .CLKOUT2(crg_clkout2),
        .CLKOUT3(crg_clkout3),
        .CLKOUT4(crg_clkout4),
        .LOCKED(crg_locked)
);


BUFG BUFG(
        .I(crg_clkout0),
        .O(crg_clkout_buf0)
);

BUFG BUFG_1(
        .I(crg_clkout1),
        .O(crg_clkout_buf1)
);

BUFG BUFG_2(
        .I(crg_clkout2),
        .O(crg_clkout_buf2)
);

BUFG BUFG_3(
        .I(crg_clkout3),
        .O(crg_clkout_buf3)
);

BUFG BUFG_4(
        .I(crg_clkout4),
        .O(crg_clkout_buf4)
);
```

Resulting SDC:
```
create_clock -period 16.6667 -waveform {0 8.33333} crg_clkout0
[...]

PR #65, opened 2020-12-15
SDC: Add include_propagated_clocks switch

tmichalak:
This PR solves issue #64 by adding the `-include_propagated_clocks` switch which adds all propagated clocks to the written SDC.


acomodi (approved):
LGTM

PR #67, opened 2020-12-15
get_ports: Trim port name

### LINKED ISSUE #62: Fix curly brackets support in get_ports commands
Many public XDC files provide `get_ports` arguments in curly brackets:
https://github.com/Digilent/digilent-xdc/blob/master/Arty-A7-35-Master.xdc#L7

Currently, support for this feature is missing / incomplete. The following XDC file:
```
set_property PACKAGE_PIN E3 [get_ports { clk }]
set_property IOSTANDARD LVCMOS33 [get_ports { clk }]
create_clock -period 10.0 clk
```
causes the following error during the Yosys synthesis step:
```
...
7.38.3. Continuing TECHMAP pass.
Using template $paramod\$lut\WIDTH=1\LUT=2'01 for cells of type $lut.
No more expansions possible.
7.39. Executing XILINX_DFFOPT pass (optimize FF control signal usage).
Optimizing FFs in top.
7.40. Executing OPT_LUT_INS pass (discard unused LUT inputs).
Optimizing LUTs in top.
Removed 0 unused cells and 2 unused wires.
8. Executing HIERARCHY pass (managing design hierarchy).
8.1. Analyzing design hierarchy..
Top module:  \top
8.2. Analyzing design hierarchy..
Top module:  \top
Removed 0 unused modules.
set_property PACKAGE_PIN E3 [get_ports { clk }]
set_property IOSTANDARD LVCMOS33 [get_ports { clk }]
create_clock -period 10.0 clk
Warning: Couldn't find port matching  clk
ERROR: set_property IO_LOC_PAIRS: Incorrect number of arguments.
```

After removing the curly brackets, everything works as intended:
```
set_property PACKAGE_PIN E3 [get_ports clk]
set_property IOSTANDARD LVCMOS33 [get_ports clk]
create_clock -period 10.0 clk
```

tmichalak:
This PR fixes #62


design_introspection-plugin/get_ports.cc
```diff
@@ -18,6 +18,7 @@
  *
  */
 #include "get_ports.h"
+#include <boost/algorithm/string.hpp>
```
litghost, line 21: Do we have other uses of `boost` in `yosys-symbiflow-plugins`?  I don't think adding boost just for `trim` makes sense.

tmichalak: We do not, but it's not something we are adding. Yosys has the boost libraries and we are building our plugins with Yosys's libraries. So we are not adding anything, we are just using something that Yosys uses already.

litghost: I'd prefer we not add this dependency.  `trim` is a pretty small function, and I'd rather not add the dependency here.  If yosys drops `boost`, I'd rather not keep it just for the `trim` function.

Please implement a `trim` function, and add unit test or two.

tmichalak: Done

common/utils.h
```diff
@@ -0,0 +1,19 @@
+#include <algorithm>
+#include <cctype>
+#include <locale>
+
+void trim_left(std::string &str)
```
litghost, line 5: Because these are in the header, they need to be marked inline.

tmichalak: Fixed.

PR #68, opened 2020-12-18
XDC: add SYN_BUFs to supported primitives

kowalewskijan:
This PR adds `SYN_IBUF` and `SYN_OBUF` support to the XDC plugin.


tmichalak (changes requested):
The changes look good, but we would need a test with the SYN_I/OBUFS

PR #72, opened 2021-01-19
Load plugins only if the symbol we're interested in is not there yet.

hzeller:
In cases the plugins are compiled into yosys already, and no shared
objects with the plugins exist, the tests would otherwise fail.

This makes it possible to have the tests behave the same whether the
plugins are compiled in or not.

### DIFF
--- a/design_introspection-plugin/tests/get_cells/get_cells.tcl
+++ b/design_introspection-plugin/tests/get_cells/get_cells.tcl
@@ -1,7 +1,6 @@
 yosys -import
-plugin -i design_introspection
-#Import the commands from the plugins to the tcl interpreter
-yosys -import
+if { [info procs get_cells] == {} } { plugin -i design_introspection }
+yosys -import  ;# ingest plugin commands
 
 read_verilog $::env(DESIGN_TOP).v
 # Some of symbiflow expects eblifs with only one module.
--- a/sdc-plugin/tests/get_clocks/get_clocks.tcl
+++ b/sdc-plugin/tests/get_clocks/get_clocks.tcl
@@ -1,8 +1,7 @@
 yosys -import
-plugin -i sdc
-plugin -i design_introspection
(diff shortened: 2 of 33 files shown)


design_introspection-plugin/tests/get_cells/get_cells.tcl
```diff
@@ -1,7 +1,6 @@
 yosys -import
-plugin -i design_introspection
-#Import the commands from the plugins to the tcl interpreter
-yosys -import
+if { [info procs get_cells] == {} } { plugin -i design_introspection }
+yosys -import  ;# ingest plugin commands
```
litghost, line 3: Please put both the plugin load and import behind the if.

hzeller: I considered that initially ,but there are a bunch of files that actually use multiple plugins, so I standardized on loading all the plugins, then finally do a `yosys -import`.

Also, it keeps the plugin load lines nice one-liners which I think makes for easier copy-paste of this boilderplate code later.

```
if { [info procs get_cells] == {} } { plugin -i design_introspection; yosys -import }
```
... would be 86 characters long..

Would there be a good reason to do `yosys -import` anyway within the `if` ?

sdc-plugin/tests/get_clocks/get_clocks.tcl
```diff
@@ -1,8 +1,7 @@
 yosys -import
-plugin -i sdc
-plugin -i design_introspection
-# Import the commands from the plugins to the tcl interpreter
-yosys -import
+if { [info procs read_sdc] == {} } { plugin -i sdc }
+if { [info procs get_nets] == {} } { plugin -i design_introspection }
+yosys -import  ;# ingest plugin commands
```
hzeller, line 4: This would be an example where we first load a bunch of plugins, then finally `yosys -import`.

PR #75, opened 2021-02-01
SDC: use fastest clock as ABC9 delay target

Ravenslofty:
This patch uses the fastest clock specified in the SDC as a delay target for ABC9's timing-driven LUT mapping, which gives it a more useful target than "fastest delay possible".

cc @kgugala


sdc-plugin/sdc.cc
```diff
@@ -181,6 +181,17 @@ struct CreateClockCmd : public Pass {
...
+
+        // Set the ABC9 delay to the shortest clock period in the design.
+        //
+        // By convention, delays in Yosys are in picoseconds, but ABC9 has
+        // no information on interconnect delay, so target half the specified
+        // clock period to give timing slack; otherwise ABC9 may produce a
+        // mapping that cannot meet the specified clock.
+        int abc9_delay = design->scratchpad_get_int("abc9.D", INT32_MAX);
+        int period_ps = period * 1000.0 / 2.0;
+        design->scratchpad_set_int("abc9.D", std::min(abc9_delay, period_ps));
```
litghost, line 193: So because `abc9.D` is a global settings, if there are more than one clock, the last clock in the design wins.  Not to mention what happens if there is a PLL/MMCM in the design.

I believe the better place to put this is at the end of the `propagate_clocks` step.  This would allow visibility into all clocks present in the design, and this call can use the minimum clock period or some other method to select between multiple clocks.

@tmichalak, thoughts?

litghost: Clock propagation happens here: https://github.com/SymbiFlow/yosys-symbiflow-plugins/blob/0a5b28fafe3d14da440291a2dc8d9af486642c84/sdc-plugin/propagation.cc#L187

There would be a need to ensure that propagation occurs before ABC9 is run.

Ravenslofty: > the last clock in the design wins

No? That's why there's a call to `std::min` here: to choose the smallest clock delay.

Ravenslofty: Further, if there's a PLL in there and the user has not described it in the SDC, then naturally the design is going to be miscompiled anyway. There's not a great deal that can be done.

mithro: @Ravenslofty - Our Yosys plugins allow propagation of clock constraints through things like PLLs (with appropriate transformations). This means that you **should** only need to specify the input clock sources.

litghost: > > the last clock in the design wins
> 
> No? That's why there's a call to `std::min` here: to choose the smallest clock delay.

Fair enough, but using clocks post propagation is likely still the right place to do this.

tmichalak: >>> the last clock in the design wins

>> No? That's why there's a call to std::min here: to choose the smallest clock delay.

> Fair enough, but using clocks post propagation is likely still the right place to do this.

If we are after the fastest clock then we should definitely do this after the clock propagation. Mostly because of what @mithro said which is that the propagation happens though PLLs. This means that the fastest clock specified in the input SDC isn't necessarily the fastest in the design.

Ravenslofty:
This now updates the delay target in `propagate_clocks` instead.

litghost (approved):
LGTM

Ravenslofty:
Hmm, this isn't working on a testcase I have.

Ravenslofty:
Nevermind; I forgot to call `propagate_clocks`.

tmichalak:
@Ravenslofty can you add a minimalistic test?

tmichalak:
@Ravenslofty please add a test so that we can proceed with merging the PR.

Ravenslofty:
My apologies for the delay, but I just added a quick test to check that the attribute is set correctly and that it's set from the minimum specified clock, not just the last specified one.

tmichalak:
> My apologies for the delay, but I just added a quick test to check that the attribute is set correctly and that it's set from the minimum specified clock, not just the last specified one.

Thanks @Ravenslofty, test looks fine, please run `make format` to fix CI failure.

Ravenslofty:
`make format` does nothing. The formatter actually seems to be deleting my SDC file.

Ravenslofty:
Ah, okay. Thank you for fixing the test.

mithro:
BTW - We should figure out why `make format` wasn't working for @Ravenslofty

tmichalak:
> BTW - We should figure out why `make format` wasn't working for @Ravenslofty

@mithro the problem was not related to `make format` not working. It was due to the fact that `make clean`, which is being run right before the format check in CI, removed the `abc9.sdc` file from the test directory. The format check simply reported this fact. All other tests use `name.input.sdc` to distinguish the test inputs files from the `.sdc` files generated during the tests.

PR #85, opened 2021-03-10
Add qlf_k4n8 plugin

kgugala:
This PR adds qlf_k4n8 synthesis flow (as discussed in https://github.com/SymbiFlow/yosys/pull/92)

This plugin is required by https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2036


tmichalak (approved):
LGTM

PR #87, opened 2021-03-12
Flip-flop reset support for QuickLogic qlf_k4n8

mkurc-ant:
This PR adds support for local flip-flop resets for the QuickLogic qlf_k4n8 architecture. It also adds an additional flip-flop type for a shift register as well as shift register inference. Tests have been updated accordingly.


kgugala:
LGTM

PR #88, opened 2021-03-17
Updating adder_lut4 INIT param as there is a change in qlf_k4n8 devic…

lnsharma:
…e adder_lut4 representation


lnsharma:
Will fix the issue, open another PR.

PR #89, opened 2021-03-17
Add formatting option for verilog files.

hzeller:
Make this a separate format-verilog target for now,
as not everyone has verible installed and there are a few
files that it can't format (because it stumbles upon
SystemVerilog keywords in the Verilog files; these files
will be automatically left alone by the formatter, so no harm
will be done).

Run [verible-verilog-format](https://github.com/google/verible/tree/master/verilog/tools/formatter) on all `*.v` files in `test/*`

Simplify double-line format call with single line
for c++ formatting.

Set editorconfig to the standard indentation
according to styleguide.

Update CI to work with current ubuntu-latest (possibly this changed recently, a few things
had to be added)
  * cmake
  * clang-format-8 (the ancient clang-format-5 doesn't exist anymore in latest ubuntu, so replace it with an slightly less ancient one. Current is clang-format-12).
  

.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         sudo apt-get install git g++-9 build-essential bison flex \
           libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot \
           pkg-config libboost-system-dev libboost-python-dev \
-          libboost-filesystem-dev zlib1g-dev clang-format-5.0
+          libboost-filesystem-dev zlib1g-dev clang-format
```
litghost, line 21: Do not use an unpinned clang-format. It means that different folks will run with different versions and they will tik-tok.

hzeller: mmh, the actual formatting information should come from the `.clang-format` file.
With a very specific .clang-format file I never had any problem.

Anyway, clang-format-11 is the current one. Should we pin to that ?

litghost: Probably `clang-format-10`, as that at least supports Ubuntu LTS (18), see https://packages.ubuntu.com/search?keywords=clang-format

litghost: Other good option would be `clang-format-8` which is on Ubuntu 16/18/20.

hzeller: Switched to clang-format-8.

tmichalak (approved):
LGTM

PR #90, opened 2021-03-18
Updating adder_lut4 INIT params as there is a change in qlf_k4n8 device lut4 representation.

lnsharma:


kgugala (approved):
LGTM, let's wait for CI

mkurc-ant (approved):
LGTM too.

PR #91, opened 2021-03-18
xdc: allow GTPE2_CHANNEL to have IO_LOC_PAIRS

acomodi:
The GTPE2_CHANNEL does not have intermediate buffers to the IOPADs. Therefore, it is necessary to allow it to have IO_LOC_PAIRS attributes to generate correct placement constraints based on the IO signals locations.

### DIFF
--- a/xdc-plugin/xdc.cc
+++ b/xdc-plugin/xdc.cc
@@ -57,7 +57,8 @@ const std::unordered_map<std::string, std::vector<std::string>> supported_primit
   {"IBUF", {"IO_LOC_PAIRS", "IOSTANDARD"}},
   {"IOBUF", {"IO_LOC_PAIRS", "IOSTANDARD", "DRIVE", "SLEW", "IN_TERM"}},
   {"IOBUFDS", {"IO_LOC_PAIRS", "IOSTANDARD", "SLEW", "IN_TERM"}},
-  {"IBUFDS_GTE2", {"IO_LOC_PAIRS"}}};
+  {"IBUFDS_GTE2", {"IO_LOC_PAIRS"}},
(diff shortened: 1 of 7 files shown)


xdc-plugin/xdc.cc
```diff
@@ -57,7 +57,8 @@ const std::unordered_map<std::string, std::vector<std::string>> supported_primit
   {"IBUF", {"IO_LOC_PAIRS", "IOSTANDARD"}},
   {"IOBUF", {"IO_LOC_PAIRS", "IOSTANDARD", "DRIVE", "SLEW", "IN_TERM"}},
   {"IOBUFDS", {"IO_LOC_PAIRS", "IOSTANDARD", "SLEW", "IN_TERM"}},
-  {"IBUFDS_GTE2", {"IO_LOC_PAIRS"}}};
+  {"IBUFDS_GTE2", {"IO_LOC_PAIRS"}},
+  {"GTPE2_CHANNEL", {"IO_LOC_PAIRS"}}};
```
litghost, line 61: This is different than the other cases, please add a nice comment.  Otherwise looks ok.

tmichalak (changes requested):
Can you add a small test for this?

xdc-plugin/xdc.cc
acomodi: Done

acomodi:
> Can you add a small test for this?

@tmichalak Done, I have also added the IBUFDS_GTE2 test which was missing as well

tmichalak:
> > Can you add a small test for this?
> 
> @tmichalak Done, I have also added the IBUFDS_GTE2 test which was missing as well

Thanks, now it all LGTM

PR #92, opened 2021-03-18
Reduce visual noise when running tests.

hzeller:
Don't clutter the output with entering/leaving directory messages or spurious warnings from tools. Have one line output per test if possible.

Warnings and other yosys logs end up in the test-specific logfile.

The test results are clearly marked and highlighted as passing/failing.

### DIFF
--- a/Makefile_test.common
+++ b/Makefile_test.common
@@ -29,7 +29,7 @@ $(1): $(1)/ok
 		touch $$<; \
 		true; \
 	else \
-		printf "Test %-18s \e[30;1mFAILED\e[0m @ %s\n" $(1) $(CURIDR); \
+		printf "Test %-18s \e[31;1mFAILED\e[0m @ %s\n" $(1) $(CURDIR); \
 		false; \
 	fi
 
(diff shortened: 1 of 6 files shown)


hzeller:
Output can be seen here: https://github.com/SymbiFlow/yosys-symbiflow-plugins/runs/2142949611?check_suite_focus=true#step:6:112

litghost:
> Output can be [seen here](https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/92/checks?check_run_id=2142949611#step:6:112)

Do I see some failures in that output, but still it returns no error?

Makefile_test.common
```diff
@@ -22,34 +22,32 @@ TEST_UTILS=../../../test-utils/test-utils.tcl
...
 
 $(1)/ok: $(1)/$(1).v
-	@echo "Running test $(1)"
 	@set +e; \
 	cd $(1); \
 	echo "source $(TEST_UTILS)" > run-$(1).tcl ;\
 	echo "source $(1).tcl" >> run-$(1).tcl ;\
 	DESIGN_TOP=$(1) TEST_OUTPUT_PREFIX=./ \
-	yosys -c "run-$(1).tcl" -q -l $(1).log; \
+	yosys -c "run-$(1).tcl" -q -q -l $(1).log; \
```
litghost, line 42: Is this supposed to be a `-q` followed by a `-q`?

tmichalak (approved):
This looks really good. I see that there are some tests that need adjustments as currently they print to stdout and mess up the one line result per test rule. But let's leave that for later.
LGTM

tmichalak:
> Do I see some failures in that output, but still it returns no error?

You probably refer to negative tests where an error message is expected.

hzeller:
@litghost there are some tests, that are actually _negative_ tests, so they are supposed to output some sort of ERROR line.

(but previously, that never was visible).

There is also some other clutter in the output in the CI currently which is from building google-test in parallel. Maybe in the CI building google test from running tests should be separated.

litghost (approved):
LGTM

PR #97, opened 2021-03-19
add support for negative edge trigfer for ffs and option to turn off …

tpagarani:
This PR adds support for negative edge clocking on FFs and also added a flag to control ff tech map.

### DIFF
--- a/ql-qlf-k4n8-plugin/synth_quicklogic.cc
+++ b/ql-qlf-k4n8-plugin/synth_quicklogic.cc
@@ -63,6 +63,9 @@ struct SynthQuickLogicPass : public ScriptPass {
         log("        By default use adder cells in output netlist.\n");
         log("        Specifying this switch turns it off.\n");
         log("\n");
+        log("    -no_ff_map\n");
+        log("        By default ff techmap is turned on. Specifying this switch turns it off.\n");
+        log("\n");
         log("\n");
         log("The following commands are executed by this synthesis command:\n");
         help_script();
@@ -72,6 +75,7 @@ struct SynthQuickLogicPass : public ScriptPass {
     string top_opt, edif_file, blif_file, family, currmodule, verilog_file;
     bool inferAdder;
     bool abcOpt;
+    bool noffmap;
 
     void clear_flags() override
     {
@@ -83,6 +87,7 @@ struct SynthQuickLogicPass : public ScriptPass {
         family = "qlf_k4n8";
         inferAdder = true;
         abcOpt = true;
+        noffmap = false;
     }
 
     void execute(std::vector<std::string> args, RTLIL::Design *design) override
@@ -119,6 +124,10 @@ struct SynthQuickLogicPass : public ScriptPass {
             }
             if (args[argidx] == "-no_abc_opt") {
                 abcOpt = false;
+                continue;
+            }
+            if (args[argidx] == "-no_ff_map") {
+                noffmap = true;
                 continue;
             }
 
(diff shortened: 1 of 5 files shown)


kgugala (changes requested):
shouldn't you add falling edge FFs to the `dfflegalize` list?  Right now it handles only P edge ones `dfflegalize -cell $_DFF_P_ x -cell $_DFF_P??_ x`

ql-qlf-k4n8-plugin/synth_quicklogic.cc
```diff
@@ -200,14 +209,16 @@ struct SynthQuickLogicPass : public ScriptPass {
...
+                run("techmap " + techMapArgs);
+            }
             run("opt_expr -mux_undef");
             run("simplemap");
             run("opt_expr");
             run("opt_merge");
             run("opt_clean");
             run("opt");
-            run("dfflegalize -cell $_DFF_P_ x -cell $_DFF_P??_ x");
+            run("dfflegalize -cell $_DFF_P_ x -cell $_DFF_P??_ x -cell $_DFF_N_ x -cell _DFF_N??_");
```
kgugala, line 221:
```suggestion
            run("dfflegalize -cell $_DFF_P_ x -cell $_DFF_P??_ x -cell $_DFF_N_ x -cell _DFF_N??_ x");
```

mkurc-ant: That's correct provided that the initial state of a `_DFF_N??_` is unknown and the flip-flop does not support initialization. If that is not the case appropriate option from the `dfflegalize` documentation should be used here. Same for other flip-flop types.

tpagarani: @kgugala @mkurc-ant looks like  _DFF_N??_ is not a valid type. where is the documentation for dfflegalize to check what types are supported?

lnsharma:
@kgugala : Can this be merged now?

kgugala:
In general the PR looks good. I'm not sure if dff INIT values are handled correctly. In the tests you set initial value with `initial q <= 1'b0;`. In the techmap INIT value is hardcoded to `1'b0`, so init from the test design is ignored.

The question here is - can the FFs have initial value? 
* If yes we should pass the value in the techmap (like e.g. here https://github.com/YosysHQ/yosys/blob/92d5550a90558a0292c8ac63cecabb2de30eb6aa/techlibs/xilinx/ff_map.v#L68).
* If no (and the FFs are always initialised to `1'b0`) we should probably add an assert in the techmap (preventing users to set init to `1'b1`)

tpagarani:
> In general the PR looks good. I'm not sure if dff INIT values are handled correctly. In the tests you set initial value with `initial q <= 1'b0;`. In the techmap INIT value is hardcoded to `1'b0`, so init from the test design is ignored.
> 
> The question here is - can the FFs have initial value?

FFs are always initialized to 0 by default. 

> 
> * If yes we should pass the value in the techmap (like e.g. here https://github.com/YosysHQ/yosys/blob/92d5550a90558a0292c8ac63cecabb2de30eb6aa/techlibs/xilinx/ff_map.v#L68).
> * If no (and the FFs are always initialised to `1'b0`) we should probably add an assert in the techmap (preventing users to set init to `1'b1`)
> 

I think @mkurc-ant already opened a PR to take care of correcting the test case. https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/96.

kgugala (approved):

kgugala:
Let's merge this and work on FFs init #96

PR #99, opened 2021-03-20
Apply 'proc' before exporting to JSON.

hzeller:
Thus all pending initial processes have been baked into
the netlist. Latest Yosys will bail when exporting JSON
otherwise.

### DIFF
--- a/sdc-plugin/tests/counter/counter.tcl
+++ b/sdc-plugin/tests/counter/counter.tcl
@@ -23,4 +23,6 @@ close $fh
 
 # Write out the SDC file after the clock propagation step
 write_sdc [test_output_path "counter.sdc"]
+
+yosys proc
 write_json [test_output_path "counter.json"]
--- a/xdc-plugin/tests/io_loc_pairs/io_loc_pairs.tcl
+++ b/xdc-plugin/tests/io_loc_pairs/io_loc_pairs.tcl
@@ -19,5 +19,6 @@ synth_xilinx -flatten -abc9 -nosrl -noclkbuf -nodsp -run prepare:check
 read_xdc -part_json [file dirname $::env(DESIGN_TOP)]/../xc7a35tcsg324-1.json $::env(DESIGN_TOP).xdc
 
 # Write the design in JSON format.
+yosys proc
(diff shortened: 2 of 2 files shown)


sdc-plugin/tests/counter/counter.tcl
```diff
@@ -23,4 +23,6 @@ close $fh
 
 # Write out the SDC file after the clock propagation step
 write_sdc [test_output_path "counter.sdc"]
+
+yosys proc
```
litghost, line 27: So all process's should've been converted at line 10.  The fact that we have some again means something else is wrong.

hzeller: Unfortunately, the Yosys error message would not tell anything about the internal name or origin of the process it can't export.

But if this is something that should've already been converted above, I have to leave it to the plugin-owner to figure that out :)

hzeller: I'll pass this then on to @tmichalak to investigate why processes have not been converted. Should this be a bug ?

litghost: Given that the original error is in a library cell-ish thing (e.g. `$paramod\FDRE\INIT=1'0`), I strongly suspect this is just a case where yosys itself generated the process post-proc.  But I'm not sure.

hzeller: it is an init process, so if `synth_xilinx` should've baked that into the netlist and removed the process then it should not be there anymore.

I could also imagine that some platforms might not need do do anything when the initialization is to all bits zero, and thus leaves the process in there ? I'll have to investigate. There were some of these things addressed in Yosys, maybe this was left out ?

I'll investigate (but low prio, probably will only have time later this evening)

litghost: So while it might be an init process, if that is a library cell, then the process is "internal", similar to the processes hidden with https://github.com/YosysHQ/yosys/pull/2679

hzeller: Yes, that is the change I remembered when I said 'some of these things addressed in Yosys'

PR #100, opened 2021-03-24
Replacing assignment with considitional assignment so that these vari…

lnsharma:
…ables can be assigned through command


tmichalak (approved):
LGTM

PR #104, opened 2021-04-14
move the ff techmap step after dfflegalize

tpagarani:
dfflegalize may generated additional dffs that need to be mapped. Hence moving the techmap after dffleagalize.


kgugala:
LGTM

PR #106, opened 2021-04-24
XDC: Support loading properties from dictionaries

### LINKED ISSUE #32: Dictionaries of properties are not handled in the XDC plugin
The XDC constraint file can contain dictionaries of properties to enable for a given port:

```
set_property -dict { LOC A1 IOSTANDARD LVCMOS33 } [get_ports {port}]
```

Currently the XDC plugin is not yet able to parse dictionaries.

carlosedp:
This change adds support to load pin properties from single-line
dictionaries.

Implementation and test-case were added. Test case mirrors `counter` case but with `-dict` properties (mixed usage in same file).

Fixes #32

Signed-off-by: Carlos de Paula <me@carlosedp.com>

### DIFF
--- a/xdc-plugin/xdc.cc
+++ b/xdc-plugin/xdc.cc
@@ -27,10 +27,12 @@
  *   Tcl interpreter.
  */
 #include "../common/bank_tiles.h"
+#include "../common/utils.h"
 #include "kernel/log.h"
 #include "kernel/register.h"
 #include "kernel/rtlil.h"
 #include "libs/json11/json11.hpp"
+#include <bits/stdc++.h>
 #include <cassert>
 
 USING_YOSYS_NAMESPACE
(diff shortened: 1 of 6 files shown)


carlosedp:
Pushed an update fixing the formatting. @tmichalak can you run the workflow?
Thanks!

xdc-plugin/xdc.cc
```diff
@@ -126,7 +130,28 @@ struct SetProperty : public Pass {
         if (design->top_module() == nullptr) {
             log_cmd_error("No top module detected\n");
         }
+        if (args.at(1) == "-dict") {
+            std::string dict_args = args.at(2).c_str();
```
tmichalak, line 134: I believe that `c_str()` is obsolete here.

xdc-plugin/xdc.cc
```diff
@@ -126,7 +130,28 @@ struct SetProperty : public Pass {
...
+            trim(dict_args);
+            std::stringstream args_stream(dict_args);
+            std::vector<std::string> tokens;
+            std::string intermediate;
+            while (getline(args_stream, intermediate, ' ')) {
+                tokens.push_back(intermediate);
+            }
+            for (long unsigned int i = 0; i < tokens.size(); i += 2) {
```
tmichalak, line 142: We should probably have a sanity check if the number of tokens is even and error out if this is not the case.

tmichalak (changes requested):
In general looks fine so thanks a lot. A few comments in.

xdc-plugin/xdc.cc
carlosedp: Done, removed.

xdc-plugin/xdc.cc
carlosedp: Added a validation. I had written the code but forgot to commit it! :)

PR #109, opened 2021-05-17
Fix CI failures

mkurc-ant:
As of version `20210510.0` of GH actions Ubuntu 20.04 image GCC 11 was added. Parts of the GCC 11 library used by clang (not sure how) make Yosys build fail. ~This PR solves that by uninstalling GCC 11 prior to Yosys compilation.~

This PR makes Yosys being built by GCC (version 9 in this case) instead of clang.

### DIFF
--- a/.github/workflows/setup.sh
+++ b/.github/workflows/setup.sh
@@ -41,6 +41,7 @@ start_section Install-Yosys
         cd ~/.local-src
         git clone https://github.com/SymbiFlow/yosys.git -b master+wip
         cd yosys
+        make config-gcc # Build Yosys using GCC
         PREFIX=$HOME/.local-bin make -j$(nproc)
         PREFIX=$HOME/.local-bin make install
         echo $(which yosys)


acomodi (commented):
Rather than removing gcc-11, isn't there the possibility to setup the system to use the correct gcc? (I see in the yml configuration that g++-9 is installed)

mkurc-ant:
The problem is that gcc-11 is already in the ubuntu-latest image provided by GH actions. At some point it was added there as indicated by the update to the relevant readme: https://github.com/actions/virtual-environments/commit/f20496e05dd68a662c8b4b169fe68f29b3afd77d#diff-6be9f9ce03ba57ca01e4ae937d5c2bb04893520fdf391ea87eb46acbc43ecdc5R17

The problem may be solved by changing the environment in the yaml but I don't know how to and whether it is possible to use a specific version of it.

acomodi:
Right, I think though that you can still set the default gcc to use for compilation, and use gcc-9. E.g.:

```
export CC=gcc-9
export CXX=g++-9
```

mkurc-ant:
I can just switch to use GCC for Yosys build. That would use the default GCC 9. I tested that and with it everything builds correctly. If you agree to that solution I can just do that.

Now it's built by clang (by default). And the problem is clang. When you run it inside the CI with `-v` it reports:
```
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/10
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/11
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/9
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/10
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/11
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

And IMO the problem is `Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/11`. I'm not sure why and how clang uses pieces of the GCC library. The only solution I found is to not have GCC 11 or more precisely libgcc-11.

acomodi:
All right, I think that we might run the test using the same compiler used when building the conda packages, at least we have consistency between the two.

Unsure which is the best option at this point, but, given that, given that we may not easily control what gcc version is used in Clang, removing gcc-11 might still be valid.

mkurc-ant:
@acomodi Hmm, I can see that conda packages are built using GCC (apparently a "conda" version of it) on Linux: https://github.com/hdl/conda-eda/blob/321475f1486f19b66f416f34ff288ab97c211124/syn/yosys/makefile-conda-config.patch#L10

So maybe we should switch to GCC (v9) here too ?

tmichalak:
I think moving to GCC (v9) is cleaner than removing clang-11.

mkurc-ant:
Ok, I'll do that

tmichalak (approved):
LGTM

PR #113, opened 2021-06-02
Running ABC optimization based on input option specified in command line

lnsharma:
…c reset support. Also adding a condition to do ABC optimization which is done by default but can be disabled by a command-line option

### DIFF
--- a/ql-qlf-plugin/synth_quicklogic.cc
+++ b/ql-qlf-plugin/synth_quicklogic.cc
@@ -243,10 +243,12 @@ struct SynthQuickLogicPass : public ScriptPass {
         }
 
         if (check_label("map_luts")) {
-            if (family == "qlf_k6n10") {
-                run("abc -lut 6 ");
-            } else {
-                run("abc -lut 4 ");
+            if(abcOpt) {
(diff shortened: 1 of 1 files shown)


ql-qlf-plugin/synth_quicklogic.cc
```diff
@@ -225,6 +225,7 @@ struct SynthQuickLogicPass : public ScriptPass {
                 run("shregmap -minlen 8 -maxlen 8");
             }
             if (family == "qlf_k6n10") {
+                run("async2sync");
```
mkurc-ant, line 228: I guess we shouldn't use the `async2sync` pass as it caused us some trouble in the past. This should be done as it is for `qlf_k4n8` with `dfflegalize` (a few lines below)

mkurc-ant (changes requested):
I'm not that familiar with the `k6n10` architecture yet but I think we should use the `dfflegalize` pass followed by appropriate techmaps as for `k4n8`.

lnsharma:
Removed async2sync pass. However, adding a check for enabling/disabling ABC optimization based on the command line option.

PR #116, opened 2021-06-07
Add ql-dsp plugin

samycharas:
This PR adds  : 
- ql-dsp plugin, which embeds a function that packs adder into QL_DSP when synthesizing a mac_unit design, it is called during synth_quicklogic multiplier stage.
Testcases for the ql-dsp functions are located in the ql-qlf folder, since the function is meant to be used during synth_quicklogic, I think it is easier to maintain to regroup all testcase under one plugin folder. I can move them to ql-dsp test folder if you prefer.
- Corrects a typo on dfflegalize call during synth_quicklogic.


samycharas:
@tpagarani @kkumar23 FYI.

ql-dsp-plugin/ql-dsp.cc
```diff
@@ -0,0 +1,164 @@
...
+        log("Map multipliers ($mul/QL_DSP) and multiply-accumulate ($mul/QL_DSP + $add)\n");
+        log("cells into ql DSP resources.\n");
+        log("Pack input registers (A, B, {C,D}), pipeline registers\n");
+        log("({F,J,K,G}, H), output registers (O -- full 32-bits or lower 16-bits only); \n");
+        log("and post-adder into into the QL_DSP resource.\n");
+        log("\n");
+        log("Multiply-accumulate operations using the post-adder with feedback on the {C,D}\n");
+        log("input will be folded into the DSP. In this scenario only, resetting the\n");
+        log("the accumulator to an arbitrary value can be inferred to use the {C,D} input.\n");
+        log("\n");
```
mkurc-ant, line 147: Each text line should begin with four space indentation

ql-qlf-plugin/synth_quicklogic.cc
```diff
@@ -185,7 +195,22 @@ struct SynthQuickLogicPass : public ScriptPass {
             run("peepopt");
             run("pmuxtree");
             run("opt_clean");
-
+            if (help_mode || (-!nodsp && family == "qlf_k6n10")) {
```
mkurc-ant, line 198: Shouldn't there be just `!nodsp`?

ql-qlf-plugin/synth_quicklogic.cc
```diff
@@ -185,7 +195,22 @@ struct SynthQuickLogicPass : public ScriptPass {
...
+                      "-D DSP_A_MINWIDTH=2 -D DSP_B_MINWIDTH=2 -D DSP_Y_MINWIDTH=11 "
+                      "-D DSP_NAME=$__MUL16X16",
+                    "(if -!nodsp)");
+                run("select a:mul2dsp", "              (if -!nodsp)");
+                run("setattr -unset mul2dsp", "        (if -!nodsp)");
+                run("opt_expr -fine", "                (if -!nodsp)");
+                run("wreduce", "                       (if -!nodsp)");
+                run("select -clear", "                 (if -!nodsp)");
+                run("ql_dsp", "                        (if -!nodsp)");
+                run("chtype -set $mul t:$__soft_mul", "(if -!nodsp)");
```
mkurc-ant, line 212: The option is named `-no_dsp` so the help should state `(if -no_dsp)` too.

ql-dsp-plugin/ql-dsp-pm.h
```diff
@@ -0,0 +1,1871 @@
+struct ql_dsp_pm {
```
mkurc-ant, line 1: I presume that this file was generated by Yosys' pattern matcher generator (pmgen). It is better to store the source .pmg file and make the generation part of the plugin build process.

mkurc-ant (changes requested):
Comments are in.

I wonder if we could integrate `ql-dsp` into `ql-qlf` so that there is only one plugin. It would then be easier to organize tests. Unless you intend to use the plugins separately.

kkumar23:
@samycharas : let us know your views on the comments from Maciej.

ql-dsp-plugin/ql-dsp.cc
samycharas: This file has been formatted using the format-check.sh script in the workflow folder, to me the indentation looks correct as it is currently, it would generate error during CI to apply this change.

ql-qlf-plugin/synth_quicklogic.cc
samycharas: Correct, I changed that

ql-dsp-plugin/ql-dsp-pm.h
samycharas: I agree, I've added the generation script to generate it during plugin build.

samycharas:
@mkurc-ant I do not intend to use both plugin separately, so I've followed your suggestion and integrated ql-dsp plugin in ql-qlf.

ql-qlf-plugin/pmgen/pmgen.py
```diff
@@ -0,0 +1,797 @@
+#!/usr/bin/env python3
```
mkurc-ant, line 1: This file is part of Yosys and certainly should not be checked into this repo. Since it is not installed along with Yosys and is needed for the build I'm not entirely sure how to proceed. Maybe add downloading it using wget directly from YosysHQ GitHub in the `pre-build` stage?

mkurc-ant (changes requested):
Looks good except for `pmgen.py` checked into this repo directly.

mkurc-ant:
@acomodi @tmichalak Could you look into this PR? It now requires Yosys' `pmgen.py` which is not installed along with Yosys to build the pattern matcher for the plugin. Do you have any other ideas how to deal with it other than manually fetching it from GitHub using eg `wget` ?

acomodi:
@mkurc-ant Another option might be to add yosys as a `third_party` submodule and directly use the `pmgen.py` script from there.

tmichalak:
@mkurc-ant @acomodi so the question here is whether `pmgen.py` will change from where you `wget` it. Making it part of third_party will ensure that won't change without somebody explicitly updating the submodule.

samycharas:
I've done the following change in the new commits :
- pmgen.py is now being downloaded from symbiflow's yosys repository
- Added generated files to .gitignore for 2 reasons, to avoid anyone to commit them, especially the pmgen.py which we are not cleaning ( the file shouldn't change after it's been downloaded, so I did not add it to clean rule ) and because they were failing format pass.
- rebased on top of the master, there should be no conflict with the branch now.
Waiting for CI.

mkurc-ant (approved):
Perfect, LGTM

PR #117, opened 2021-06-21
Fixing an issue in edf file generation where hilomap is required but …

lnsharma:
…it was not defined. Also commenting an include file which is causing compilation error in mac compilation and commenting it has no impact on Ubuntu compilation

### DIFF
--- a/xdc-plugin/xdc.cc
+++ b/xdc-plugin/xdc.cc
@@ -32,7 +32,6 @@
 #include "kernel/register.h"
 #include "kernel/rtlil.h"
 #include "libs/json11/json11.hpp"
-//#include <bits/stdc++.h>
 #include <cassert>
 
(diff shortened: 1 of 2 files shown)


xdc-plugin/xdc.cc
```diff
@@ -32,7 +32,7 @@
 #include "kernel/register.h"
 #include "kernel/rtlil.h"
 #include "libs/json11/json11.hpp"
-#include <bits/stdc++.h>
+//#include <bits/stdc++.h>
```
mkurc-ant, line 35: If the CI passes with this include commented out I'm ok with removing it.

mkurc-ant (approved):
Please remove the commented include line completely. Otherwise LGTM

mkurc-ant (approved):
ok, LGTM

PR #120, opened 2021-07-02
Adding support for writing EQN param in edif file. Also, removing sup…

lnsharma:
…port to write INIT param for DFFs

### DIFF
--- a/ql-qlf-plugin/Makefile
+++ b/ql-qlf-plugin/Makefile
@@ -1,10 +1,10 @@
-# Copyright (C) 2020-2021  The SymbiFlow Authors.
+#Copyright(C) 2020 - 2021 The SymbiFlow Authors.
 #
-# Use of this source code is governed by a ISC-style
-# license that can be found in the LICENSE file or at
-# https://opensource.org/licenses/ISC
+#Use of this source code is governed by a ISC - style
+#license that can be found in the LICENSE file or at
+#https: // opensource.org/licenses/ISC
 #
-# SPDX-License-Identifier:ISC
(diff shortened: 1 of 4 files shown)


ql-qlf-plugin/Makefile
```diff
@@ -9,7 +9,8 @@
 NAME = ql-qlf
 SOURCES = synth_quicklogic.cc \
           ql-dsp.cc \
-          pp3_braminit.cc
+          pp3_braminit.cc \
+	  quicklogic_eqn.cc
```
acomodi, line 13: This seems to be misaligned with the rest

acomodi (changes requested):
There is one minor formatting issue, but otherwise LGTM

ql-qlf-plugin/Makefile
lnsharma: Corrected it

ql-qlf-plugin/Makefile
```diff
@@ -1,15 +1,17 @@
...
 #
-# Use of this source code is governed by a ISC-style
-# license that can be found in the LICENSE file or at
-# https://opensource.org/licenses/ISC
+#Use of this source code is governed by a ISC - style
+#license that can be found in the LICENSE file or at
+#https: // opensource.org/licenses/ISC
 #
-# SPDX-License-Identifier:ISC
+#SPDX - License - Identifier : ISC
```
acomodi, line 7: I think this was changed by the clang formatter. This change should be reverted

acomodi (approved):
LGTM, merging

PR #121, opened 2021-07-16
Removing dff absorption operated by ql_dsp

samycharas:
This PR removes subpattern matching for dff cells performed during ql_dsp pass.
Currently QL_DSP performs strictly combinatorial operations, and ff absorption in some testcases breaks the synthesis.
dff subpattern have been commented out instead of being removed because we will use them in the future when QL_DSP will be converted into a MAC Unit.

Signed-off-by: Samy <samy.charas@cpe.fr>


mkurc-ant:
@samycharas I've verified that with this change Yosys now handles `QL_DSP` outputs correctly.

mkurc-ant:
@lnsharma @tpagarani Are you ok with the changes in this PR ?

tpagarani:
@mkurc-ant , yes these changes should be fine

PR #122, opened 2021-07-19
Fix adder mapping

samycharas:
This PR modifies adders mapping for the qlf_k6n10 device.
The carry chain input is now longer directly connected to the first adder in chain CIN, but is generated using an intermediate adder.

This PR also reference ql_dsp function as part of the ql-qlf plugin in the main readme, and fix a typo qlf_k6n10 cells sim file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ The plugin adds the following command:
 
 The plugin adds the following command:
 * synth_quicklogic
+* ql_dsp
 
 Detailed help on the supported command(s) can be obtained by running `help <command_name>` in Yosys.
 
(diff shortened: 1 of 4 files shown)


tpagarani:
@acomodi @mkurc-ant can you please merge this PR after review

acomodi (changes requested):
This PR requires a rebase. In addition, after https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/127, all the library files for each arch do not have the arch prefix, as they are being installed in different directories.

acomodi (approved):
LGTM

PR #123, opened 2021-07-26
Rectifying the port name as 'a' instead of 'A' for logic_0 and logic_…

lnsharma:
…1 for pp3 device family


mkurc-ant (approved):
LGTM. Please note that this requires a corresponding change to SymbiFlow (PP3 techmaps).

PR #124, opened 2021-07-26
PP3e ASSP[LR] support

mkurc-ant:
This PR adds definitions of left and right ASSP cells of PP3e (as blackboxes). Cell names and port definitions were derived from https://github.com/QuickLogic-Corp/PolarPro3E/blob/master/Device%20Architecture%20Files/QLAL3S2_lib.xml


mkurc-ant:
@lnsharma @tpagarani Could you please verify that the ASSP cell definitions are correct?

lnsharma (approved):
LGTM

PR #126, opened 2021-08-12
Introduce splitports plugin

lpawelcz:
This PR adds the `splitports` plugin which makes sure that every signal in EDIF file would by a single bit one instead of an array

Although the core logic for splitting the signals is working, this is still a WIP since there are some bugs that need to be fixed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 #
 # SPDX-License-Identifier:ISC
 
-PLUGIN_LIST := fasm xdc params sdc ql-iob design_introspection integrateinv ql-qlf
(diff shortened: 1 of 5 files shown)


mithro:
I was sure this functionality was already in upstream Yosys?

lpawelcz:
There is a `splitnets` command already in Yosys. It splits the multi-bit nets into single-bit ones.
However there is an use case when `EDIF` file is required to have only single bit ports in it but `splitnets -ports` command does not do the job correctly.
This plugin is meant to do split in such a way, that every port in the EDIF would be single bit

lpawelcz:
I'm closing this because this issue was solved with custom EDIF writer in: https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/128

PR #131, opened 2021-08-25
ci: use upstream yosys and add ccache in GH actions

acomodi:


mithro:
Awesome!

PR #143, opened 2021-11-10
ql-qlf: add possibility to override synth pass name through define

acomodi:

This PR adds the possibility to override the `synth_quicklogic` pass name when building the plugin.

The pass name can be overriden by setting an extra flag as follows:
```
make EXTRA_FLAGS="-DPASS_NAME=new_name"
```


kgugala (approved):
LGTM

PR #146, opened 2021-11-16
Adding -nodffe -nosdff to opt pass in case of qlf_k4n8 and qlf_k6n10 …

lnsharma:
…device, so as to avoid mapping to DFFE or SDFF


mkurc-ant (approved):
LGTM

PR #151, opened 2021-11-23
Fix packed ranges in parameter

kamilrakoczy:
This PR fixes packed ranges in parameters using mainline yosys.

### DIFF
--- a/uhdm-plugin/UhdmAst.cc
+++ b/uhdm-plugin/UhdmAst.cc
@@ -1237,7 +1237,7 @@ void UhdmAst::convert_packed_unpacked_range(AST::AstNode *wire_node, const std::
     size_t packed_size = 1;
     size_t unpacked_size = 1;
     std::vector<AST::AstNode *> ranges;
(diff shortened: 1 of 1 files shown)


uhdm-plugin/UhdmAst.cc
```diff
@@ -2659,7 +2659,15 @@ void UhdmAst::process_parameter()
         case vpiArrayTypespec: {
             shared.report.mark_handled(typespec_h);
 #ifdef BUILD_UPSTREAM
-            visit_one_to_many({vpiRange}, typespec_h, [&](AST::AstNode *node) { packed_ranges.push_back(node); });
+            visit_one_to_one({vpiElemTypespec}, typespec_h, [&](AST::AstNode *node) {
+                if (node) {
```
kbieganski, line 2663: Why not combine these two `if`s?

kbieganski (changes requested):

uhdm-plugin/UhdmAst.cc
kamilrakoczy: Fixed

PR #158, opened 2021-12-07
K6n10 dsp

tpagarani:
Added a blackbox DSP model for K6N10f


tpagarani:
@mkurc-ant I created a new PR with single commit, https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/159

Issue #144, opened 2021-11-12
Wrong synthesis result when buses are not 0-addressed


gergoerdi:
This is in relation to https://github.com/YosysHQ/yosys/issues/3078 where I was told that the debug output of Yosys is just misleading, i.e. the problem should be elsewhere in the toolchain.

Given the following Verilog input:

```
module Top(
           output [5:2] LED
           );

   assign LED[5:2] = 4'b1101;
endmodule
```

I am synthesizing for the Nexys A7 50T dev board with a Xilinx xc7a50tcsg324 FPGA. I connect `LED[5:2]` to the LEDs on board with the following XDC:

```
set_property -dict { PACKAGE_PIN J13   IOSTANDARD LVCMOS33 } [get_ports { LED[2] }];
set_property -dict { PACKAGE_PIN N14   IOSTANDARD LVCMOS33 } [get_ports { LED[3] }];
set_property -dict { PACKAGE_PIN R18   IOSTANDARD LVCMOS33 } [get_ports { LED[4] }];
set_property -dict { PACKAGE_PIN V17   IOSTANDARD LVCMOS33 } [get_ports { LED[5] }];
```

If I synthesize this with Vivado, LEDs 2, 4 and 5 light up, as expected. If I synthesize this with Symbiflow, instead only LED 2 lights up. 

Full reproducer is at https://github.com/gergoerdi/yosys-symbiflow-plugins-issue-144/tree/81f4cd6878574614f08269e44ec8df72a857c640

mithro:
I'm pretty sure this is an issue with the XDC plugin as a number of the `OBUFT_VPR` modules do not have `IO_LOC_PAIRS` value set.

mithro:
PTAL - @tmichalak

tmichalak:
Yes, this looks like a bug in the plugin. I will look into it.

tmichalak:
Closed with #161

PR #161, opened 2021-12-10
xdc: fix port numbering

lpawelcz:
This PR fixes a bug in XDC plugin that was described in https://github.com/SymbiFlow/yosys-symbiflow-plugins/issues/144
`port_bit` returned by `extract_signal()` is the same as index of given port wire, while `chunk.offset`s are always indexed starting from 0. That causes inconsistency when ports of the module are not 0-indexed as in example:
```
module Top(
           output [5:2] LED
           );
```
This required correcting `port_bit` with `start_offset` of the whole port.


tmichalak (approved):
LGTM

Issue #177, opened 2021-12-29
ERROR: Incorrect period value


fontamsoc:
https://lists.librecores.org/pipermail/symbiflow/2021-December/000049.html

Hello,

I am attempting to build [FonTamSOC](https://github.com/fontamsoc)
targeting NexysA7.

However, `symbiflow_synth` fails with `ERROR: Incorrect period value`
at the end of the logs.

Steps to reproduce failure:
```
# clone verilog sources:
git clone https://github.com/fontamsoc/hw.git
cd hw

# prerequisite:
ln -snf pu32-nexys4ddr/litedram/litedram.hex litedram.hex

# synthesize:
symbiflow_synth -t nexys4ddr -v pu32-nexys4ddr/nexys4ddr.v -d artix7
-p xc7a100tcsg324-1 -x pu32-nexys4ddr/nexys4ddr.xdc &>/dev/null

tail nexys4ddr_synth.log
```

> The error seems to come from
> https://github.com/SymbiFlow/yosys-symbiflow-plugins/blob/master/sdc-plugin/sdc.cc#L153
> 
> I'm guessing that the error is coming from line 4 of your .xdc file but the
> current output doesn't really give you enough information to know or
> confirm that. I recommend logging a bug on either symbiflow-examples or
> yosys-symbiflow-plugins with your reproduction instructions and output you
> have provided.
> 
> Next steps would probably be to improve the error message with the value it
> thinks is invalid. The code also doesn't seem to know the difference
> between the create_clock command getting invalid value for -period and the
> command not getting a -period argument all together.
> 
> Second step would be to figure out how to create / throw the error message
> so it reports the file and line number where the issue is found.
> 
> Hope that helps,
> 
> Tim 'mithro' Ansell

tmichalak:
@fontamsoc I couldn't reproduce the error on my end. What version of the symbiflow-plugins do you use? Is it the same as [here](https://github.com/SymbiFlow/symbiflow-examples/blob/master/xc7/environment.yml#L11) ?
Have you ever built the sdc plugin manually. If so then there is a [debug option](https://github.com/SymbiFlow/yosys-symbiflow-plugins/blob/master/Makefile_plugin.common#L51) that prints out additional information during the clock propagation phase.
Also, should I use the code on the master branch or some other commit, because the `44c4298` SHA that is in the report can't be found.

fontamsoc:
HI @tmichalak,

[...]

PR #205, opened 2022-01-27
sdc: Allow -add option in the create_clock command

tmichalak:
This PR fixes the problem reported in #177 without having to remove the `-add` option from the `create_clock` command.


mkurc-ant (approved):
LGTM

PR #212, opened 2022-02-04
Adds class struct variable support.

QuantamHD:
Signed-off-by: Ethan Mahintorabi <ethanmoon@google.com>

### DIFF
--- a/uhdm-plugin/UhdmAst.cc
+++ b/uhdm-plugin/UhdmAst.cc
@@ -3597,6 +3597,7 @@ AST::AstNode *UhdmAst::process_object(vpiHandle obj_handle)
     case vpiEnumVar:
     case vpiEnumNet:
     case vpiStructVar:
+    case vpiClassVar:
     case vpiStructNet:
         process_custom_var();
         break;


QuantamHD:
Comes up in situations like the following

```
// mystruct::the_struct declaration omitted

function automatic mystruct::the_struct do_stuff_function;
  input mystruct::the_struct a;
  begin
    do_stuff_function = a
  end
endfunction
```

QuantamHD:
The class_var was removed from our UHDM with this change in Surelog https://github.com/chipsalliance/Surelog/pull/2539, that was found when I reported https://github.com/chipsalliance/Surelog/issues/2537.

kamilrakoczy:
Looks like this is issue in the Surelog itself. I've created issue in Surelog for this: https://github.com/chipsalliance/Surelog/issues/2550

I've tested example that you reported in: https://github.com/chipsalliance/Surelog/issues/2537, using typedefs directly from package gives ``vpiClassVar`` as variable type, but it should give ``vpiStructVar``, but if we import whole package, and then use typedefs, Surelog gives us correct ``vpiStructVar``, e.g.:
```
package b;
  typedef struct packed {
    logic                one;
  } foo_bar_t;

endpackage : b

package aa;
  import b::*;
  function automatic foo_bar_t foo_bar;
    foo_bar_t y;

  endfunction : foo_bar
endpackage : aa
```

kbieganski (dismissed):

hzeller:
Is this resolved ?

PR #222, opened 2022-02-09
Skip simulation functions

RRozak:


kbieganski (approved):
LGTM

uhdm-plugin/UhdmAst.cc
```diff
@@ -3123,24 +3123,26 @@ void UhdmAst::process_sys_func_call()
...
     } else if (current_node->str == "\\$unsigned") {
         current_node->type = AST::AST_TO_UNSIGNED;
-    } else if (current_node->str == "\\$display" || current_node->str == "\\$time") {
-        current_node->type = AST::AST_TCALL;
-        current_node->str = current_node->str.substr(1);
     } else if (current_node->str == "\\$readmemh") {
         current_node->type = AST::AST_TCALL;
     }
 
+    std::string simulation_functions[] = {
```
kamilrakoczy, line 3130: Some of this functions are already skipped in the yosys: https://github.com/YosysHQ/yosys/blob/master/frontends/ast/simplify.cc#L935

Where others like ``display`` or ``write`` are partially implemented: https://github.com/YosysHQ/yosys/blob/master/frontends/ast/simplify.cc#L942

I'd say we shouldn't skip functions that are also skipped on the yosys side or those partially implemented (especially ``display`` as it is used in ``ibex``/``opentitan``)

kamilrakoczy (changes requested):

PR #226, opened 2022-02-11
qlf_k6n10f DSP multiplier inference

mkurc-ant:
This PR adds preliminary support for inference of DSP-based multipliers for `qlf_k6n10f`.

There are two new blackbox models `dsp_t1_20x18x64` and `dsp_t1_10x9x32` representing the whole DSP and one half of it. Simulation models for them will be added in subsequent PRs. Also their definitions (ports/parameters) may change.

The `synth_quicklogic` pass for  `qlf_k6n10f` now has the ability to infer multipliers. This is done by the generic `mul2dsp.v` techmap from Yosys. Techmapping is done in multiple stages where each subsequent stage maps to "smaller" DSPs. The techmap transforms internal `$mul` cells into `__QL_MUL20X18` or `__QL_MUL10X9` which are then techmapped to the final DSP cells present in the architecture by `qlf_k6n10f/dsp_map.v`.

There is a test added which verifies if the inference takes place but since there are no simulation models yet there is no equivalence checking / simulation.

### DIFF
--- a/ql-qlf-plugin/qlf_k6n10f/cells_sim.v
+++ b/ql-qlf-plugin/qlf_k6n10f/cells_sim.v
@@ -664,22 +664,61 @@ module TDP_BRAM36 (
 
 endmodule
 
-(* blackbox *)
-module QL_DSP1 (
-    input  [19:0] a,
-    input  [17:0] b,
-    input  clk0,
-    (* clkbuf_sink *)
-    input  clk1,
-    (* clkbuf_sink *)
-    input  [ 1:0] feedback0,
-    input  [ 1:0] feedback1,
-    input  load_acc0,
-    input  load_acc1,
-    input  reset0,
-    input  reset1,
-    output reg [37:0] z
-);
-    parameter MODE_BITS = 27'b00000000000000000000000000;
-endmodule  /* QL_DSP1 */
+//(* blackbox *)
+//module QL_DSP1 (
+//    input  [19:0] a,
+//    input  [17:0] b,
+//    input  clk0,
+//    (* clkbuf_sink *)
+//    input  clk1,
+//    (* clkbuf_sink *)
+//    input  [ 1:0] feedback0,
+//    input  [ 1:0] feedback1,
+//    input  load_acc0,
+//    input  load_acc1,
+//    input  reset0,
+//    input  reset1,
+//    output reg [37:0] z
+//);
+//    parameter MODE_BITS = 27'b00000000000000000000000000;
+//endmodule  /* QL_DSP1 */
 
+(* blackbox *) // TODO: add sim model
+module dsp_t1_20x18x64 (
+    input  [63:0] a_i,
+    input  [17:0] b_i,
+    output [63:0] z_o,
+
+    input         clock_i,
+    input         reset_i,
+    input         load_acc_i,
+
+    input         register_inputs_i,
+    input         subtraction_mode_i,
+    input  [1:0]  feedback_i,
+    input         round_i,
+    input  [5:0]  shift_right_i,
+    input         saturate_enable_i,
+    input  [1:0]  output_select_i
+);
+endmodule
+
+(* blackbox *) // TODO: add sim model
+module dsp_t1_10x9x32 (
+    input  [31:0] a_i,
+    input  [ 8:0] b_i,
+    output [31:0] z_o,
+
+    input         clock_i,
+    input         reset_i,
+    input         load_acc_i,
+
+    input         register_inputs_i,
+    input         subtraction_mode_i,
+    input  [1:0]  feedback_i,
+    input         round_i,
+    input  [5:0]  shift_right_i,
+    input         saturate_enable_i,
+    input  [1:0]  output_select_i
+);
+endmodule
--- a/ql-qlf-plugin/qlf_k6n10f/dsp_map.v
+++ b/ql-qlf-plugin/qlf_k6n10f/dsp_map.v
@@ -6,22 +6,77 @@
 //
 // SPDX-License-Identifier:ISC
 
-module \$__MUL16X16 (input [15:0] A, input [15:0] B, output [31:0] Y);
-	parameter A_SIGNED = 0;
-	parameter B_SIGNED = 0;
-	parameter A_WIDTH = 0;
-	parameter B_WIDTH = 0;
-	parameter Y_WIDTH = 0;
+module \$__QL_MUL20X18 (input [19:0] A, input [17:0] B, output [37:0] Y);
+    parameter A_SIGNED = 0;
+    parameter B_SIGNED = 0;
+    parameter A_WIDTH = 0;
+    parameter B_WIDTH = 0;
+    parameter Y_WIDTH = 0;
 
-	QL_DSP #(
-		.A_REG(1'b0),
-		.B_REG(1'b0),
-		.C_REG(1'b0),
-		.D_REG(1'b0),
-		.ENABLE_DSP(1'b1),
-	) _TECHMAP_REPLACE_ (
-		.A(A),
-		.B(B),
-		.O(Y),
-	);
+    wire [19:0] a;
+    wire [17:0] b;
+    wire [63:0] z;
+
+    assign a = (A_WIDTH == 20) ? A :
+               (A_SIGNED) ? {{(20 - A_WIDTH){A[A_WIDTH-1]}}, A} :
+                            {{(20 - A_WIDTH){1'b0}},         A};
+
+    assign b = (B_WIDTH == 18) ? B :
+               (B_SIGNED) ? {{(18 - B_WIDTH){B[B_WIDTH-1]}}, B} :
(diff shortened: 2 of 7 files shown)


mkurc-ant:
@tpagarani Please review

ql-qlf-plugin/qlf_k6n10f/cells_sim.v
```diff
@@ -664,22 +664,61 @@ module TDP_BRAM36 (
...
-    input  clk1,
-    (* clkbuf_sink *)
-    input  [ 1:0] feedback0,
-    input  [ 1:0] feedback1,
-    input  load_acc0,
-    input  load_acc1,
-    input  reset0,
-    input  reset1,
-    output reg [37:0] z
+//(* blackbox *)
```
tpagarani, line 667: Leave QL_DSP1 block in there for now as we have 2 different versions of DSP blocks in k6n10. Right now we won't target the inference for the older DSP block

ql-qlf-plugin/qlf_k6n10f/cells_sim.v
```diff
@@ -664,22 +664,61 @@ module TDP_BRAM36 (
...
-endmodule  /* QL_DSP1 */
+endmodule
 
+(* blackbox *) // TODO: add sim model
+module dsp_t1_10x9x32 (
+    input  [31:0] a_i,
+    input  [ 8:0] b_i,
+    output [31:0] z_o,
+
+    input         clock_i,
```
tpagarani, line 712: add (* clkbuf_sink *) attribute to clock_i

ql-qlf-plugin/qlf_k6n10f/dsp_map.v
```diff
@@ -6,22 +6,77 @@
...
+    assign b = (B_WIDTH == 18) ? B :
+               (B_SIGNED) ? {{(18 - B_WIDTH){B[B_WIDTH-1]}}, B} :
+                            {{(18 - B_WIDTH){1'b0}},         B};
+
+    dsp_t1_20x18x64 _TECHMAP_REPLACE_ (
+        .a_i    (a),
+        .b_i    (b),
+        .z_o    (z),
+
+        .register_inputs_i  (1'b0),
```
tpagarani, line 33: all the configurable inputs have to be modeled as MODE_BITS. Please refer to this for configurable bits https://github.com/QuickLogic-Corp/factor/blob/main/DSP/dsp_type1_bw/rtl/dsp.sv#L28-L36

ql-qlf-plugin/qlf_k6n10f/dsp_map.v
```diff
@@ -6,22 +6,77 @@
...
+
+    assign b = (B_WIDTH ==  9) ? B :
+               (B_SIGNED) ? {{( 9 - B_WIDTH){B[B_WIDTH-1]}}, B} :
+                            {{( 9 - B_WIDTH){1'b0}},         B};
+
+    dsp_t1_10x9x32 _TECHMAP_REPLACE_ (
+        .a_i    (a),
+        .b_i    (b),
+        .z_o    (z),
+
```
tpagarani, line 69: all the configurable inputs have to be modeled as MODE_BITS. Please refer to this for configurable bits https://github.com/QuickLogic-Corp/factor/blob/main/DSP/dsp_type1_bw/rtl/dsp.sv#L28-L36

ql-qlf-plugin/qlf_k6n10f/dsp_map.v
```diff
@@ -6,22 +6,77 @@
...
-		.C_REG(1'b0),
-		.D_REG(1'b0),
-		.ENABLE_DSP(1'b1),
-	) _TECHMAP_REPLACE_ (
-		.A(A),
-		.B(B),
-		.O(Y),
-	);
+module \$__QL_MUL20X18 (input [19:0] A, input [17:0] B, output [37:0] Y);
+    parameter A_SIGNED = 0;
```
tpagarani, line 10: How do B_SIGNED and A_SIGNED get set?

mkurc-ant: Those are initially set for `$mul` cells by the `alumacc` pass depending on whether wires for signals being multiplied are declared as signed or not. For this cell type they are propagated by the generic `dsp_map.v` techmap from the original `$mul` cell.

ql-qlf-plugin/qlf_k6n10f/cells_sim.v
mkurc-ant: @tpagarani Alright. My question is: should we techmap `dsp_t1_20x18x64` and `dsp_t1_10x9x32` to `QL_DSP1` at the end of the flow so that the final netlist contains only `QL_DSP1` ?.

The problem with that is `QL_DSP1` represents the fractuable DSP as a whole which means that we cannot "pack" two `dsp_t1_10x9x32` into it at the synthesis stage. For that the FPGA arch should be updated to have two modes for the DSP - one for the whole thing and one for two halves as individual sites.

ql-qlf-plugin/qlf_k6n10f/cells_sim.v
mkurc-ant: Added.

ql-qlf-plugin/qlf_k6n10f/dsp_map.v
mkurc-ant: I managed to find correspondence between the those control ports and `MODE_BITS` there: https://github.com/QuickLogic-Corp/australis/blob/cb301e8806fe7513d4a87bb9f56116d547296e06/IPs/eFPGA/arch-variants/K6N10-B/tsmc22/CustomModules/ql_dsp.v#L78-L86. Not sure though if this is the correct place.

My intention for introduction of `dsp_t1_20x18x64` and `dsp_t1_10x9x32` cells was that they are the ones that can be either explicitly instantiated by a user or inferred by Yosys. Then In the end of the flow they are mapped to `QL_DSP1` and `MODE_BITS` are set in the final techmap basing either on control port connections (to consts) or individual control parameters.

The problem here is `QL_DSP1` corresponds to the full DSP so we cannot map `dsp_t1_10x9x32` (half DSP) to it at the synthesis stage.

The best solution here would be to add support for a cell like "`QL_DSP1_HALF`" to the architecture along with another mode for the DSP pb_type in VPR. Then `QL_DSP1` will have `MODE_BITS[26:0]` parameter mapped directly to config bits while `QL_DSP1_HALF` would have MODE_BITS[12:0] mapped to the corresponding half of it.

From what I can tell initially `MODE_BITS` contains 1-bit that controls fractuation plus 2x13-bit words controlling each DSP half. I presume that both DSP halves are intependent in terms of control bits. If not then the situation gets complicated as VPR would need to know that it can pack together only DSPs with matching shared part of their configs.

tpagarani: @mkurc-ant, mapping of mode bits for this DSP are located https://github.com/QuickLogic-Corp/australis/blob/main/IPs/eFPGA/arch-variants/K6N10_1212_DSP/tsmc22/CustomModules/ql_dsp.v
The control port in fracturable mode are common as the DSP operates as SIMD in this case.
Is there a way to map SIMD instructions in yosys? I do see some experimental work in "xilinx_simd_pack"

tpagarani: @mkurc-ant on a second thought, VPR does support packing multiple FFs in a CLB only if their control pins (Set/Reset/Clock/Enable) are identical. So why can't the same approach be used for dsp_t1_10x9x32

mkurc-ant: @tpagarani We could if those control signals were routable. But unfortunately they are just static control parameters and VPR does not check cell parameter during packing. I'll look at SIMD operation inference in Yosys.

mkurc-ant:
@tpagarani I have updated definitions of DSP black-boxes. Please take a look.

I have also noticed that a new FIR filter mode with 4 fixed coeffs has been added. I updated ports of the black-boxes accordingly and I am proceeding with updating the simulation model.

tpagarani (commented):
@mkurc-ant if we need to handle different timing for different modes of operation e.g. mult/multadd/macc/saturaion etc, Can we support it with single blackbox or we need to create multiple primitive types and have a blackbox corresponding to each of the functional modes where timing data needs to be different?

mkurc-ant:
@tpagarani Hmm, from the VPR perspective cell timings are bound to pb_type. So if you want different timings for different modes of DSP operation then you need a pb_type with multiple modes in the architecture. We handled it this way for EOS-S3 RAMs. Since you are using OpenFPGA I think it means having multiple physical modes but I'm not sure.

tpagarani:
@mkurc-ant let me discuss with OpenFPGA owners. I think for now this PR is good to merge.

mkurc-ant:
@tpagarani Alright, I'm merging this PR.

PR #228, opened 2022-02-11
Add vpiForever handling

RRozak:
There is no AST::AST_FOREVER type in yosys.
I handled vpiForever in the same way as in verilator, by by creating while loop with true condition.

### DIFF
--- a/uhdm-plugin/UhdmAst.cc
+++ b/uhdm-plugin/UhdmAst.cc
@@ -3607,6 +3607,27 @@ void UhdmAst::process_while()
     });
 }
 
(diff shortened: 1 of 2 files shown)


kamilrakoczy (commented):
Have you tested this change in yosys? I think this can result in infinite loop while unrolling the loop in ``simplify`` step.

RRozak:
I tested it on this test: https://github.com/chipsalliance/UHDM-integration-tests/blob/master/tests/Forever/top.sv
I will close this PR, because currently the `forever` loop doesn't have practical usage in yosys.

PR #243, opened 2022-02-22
Skip non-synthesizable objects

kamilrakoczy:


kamilrakoczy:
This PR requires fix for infinite loop when comparing ``Class``/``Task`` objects in UHDM: https://github.com/chipsalliance/UHDM/issues/652

PR #250, opened 2022-02-24
Fixed adder techmap to map Carry out correctly

tpagarani:
Signed-off-by: Tarachand Pagarani <tpagarani@quicklogic.com>


kgugala (approved):
LGTM

PR #251, opened 2022-02-24
Fix default signed value for parameters

kamilrakoczy:

### DIFF
--- a/uhdm-plugin/UhdmAst.cc
+++ b/uhdm-plugin/UhdmAst.cc
@@ -1086,15 +1086,28 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
         // so we are treating here UInt in the same way as if they would be Int
         case vpiUIntVal:
         case vpiIntVal: {
-            auto size = vpi_get(vpiSize, obj_h);
+            int size = -1;
+            bool is_signed = false;
+            // Surelog sometimes report size as part of vpiTypespec (e.g. int_typespec)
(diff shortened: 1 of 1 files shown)


uhdm-plugin/UhdmAst.cc
```diff
@@ -1086,15 +1086,26 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
...
         case vpiUIntVal:
         case vpiIntVal: {
-            auto size = vpi_get(vpiSize, obj_h);
+            int size = -1;
+            bool is_signed = false;
+            visit_one_to_one({vpiTypespec}, obj_h, [&](AST::AstNode *node) {
+                if (node && node->attributes.count(UhdmAst::packed_ranges()) && node->attributes[UhdmAst::packed_ranges()]->children.size() &&
+                    node->attributes[UhdmAst::packed_ranges()]->children[0]->children.size()) {
+                    size = node->attributes[UhdmAst::packed_ranges()]->children[0]->children[0]->integer;
+                }
```
RRozak, line 1095: Could you add a comment describing what is done here?
It's hard to guess what these objects are, for example `node->attributes[UhdmAst::packed_ranges()]->children[0]->children[0]`

RRozak (changes requested):

uhdm-plugin/UhdmAst.cc
kamilrakoczy: Added.

RRozak (approved):

PR #252, opened 2022-02-25
k6n10f dsp macc inference

mkurc-ant:
This PR adds the `ql_dsp_macc` pass which infers multiply-accumulate (MACC) operations and implements them using `k6n10f` DSP blocks.

The inference supports MACC with output either from before or after the accumulator register. `k6n10f` DSP does not allow immediate accumulator load but allows for breaking the accumulation loop on demand. This feature is supported as well. 

The tests located under `ql-qlf-plugin/tests/qlf_k6n10f/dsp_macc` show all possible variants of how a MACC operation can be expressed in Verliog for it to be inferred correctly.

The PR also includes fixes to the plugin build system so that `pmgen` is correctly used as a dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,9 @@ endef
 
 $(foreach plugin,$(PLUGIN_LIST),$(eval $(call install_plugin,$(plugin))))
 
+pmgen.py:
+	wget -nc -O $@ https://raw.githubusercontent.com/YosysHQ/yosys/master/passes/pmgen/pmgen.py
+
 plugins: $(PLUGINS)
 
 install: $(PLUGINS_INSTALL)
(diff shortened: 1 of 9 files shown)


mkurc-ant:
@tpagarani Please take a look.

mkurc-ant:
@tpagarani That was the first approach I tried but it turned out that `$macc` does not explicitly represent a multiply-accumulate block with an integrated accumulator register. That's why even though if we used the `alumacc` pass to infer `$macc` there would still be the need for a pass that would convert it and the following `$adff` / `$dff` to the final DSP cell. A simple techmap would not be sufficient.

tpagarani (approved):

mkurc-ant:
Ok, merging.

PR #256, opened 2022-03-01
Add support for vpiMinTypMaxOp

kamilrakoczy:

### DIFF
--- a/uhdm-plugin/UhdmAst.cc
+++ b/uhdm-plugin/UhdmAst.cc
@@ -2589,6 +2589,14 @@ void UhdmAst::process_operation()
             current_node = nullptr;
             break;
         }
+        case vpiMinTypMaxOp: {
+            // ignore min and max and set only typ
+            log_assert(current_node->children.size() == 3);
+            auto tmp = current_node->children[1]->clone();
+            delete current_node;
+            current_node = tmp;
+            break;
+        }
         default: {
             delete current_node;
             current_node = nullptr;


uhdm-plugin/UhdmAst.cc
```diff
@@ -2589,6 +2589,15 @@ void UhdmAst::process_operation()
             current_node = nullptr;
             break;
         }
+        case vpiMinTypMaxOp: {
+            // ignore min and max and set only typ
+            log_assert(current_node->children.size() == 3);
+            auto tmp = current_node->children[1];
+            current_node->children.clear();
```
RRozak, line 2596: Shouldn't we delete the nodes children[0] and children[2]?

RRozak (changes requested):

uhdm-plugin/UhdmAst.cc
kamilrakoczy: True, now I'm cloning children[1] and using ``delete_children`` to make sure we delete all children: https://github.com/YosysHQ/yosys/blob/master/frontends/ast/ast.cc#L267

RRozak: `delete children` is called by AstNode destructor: https://github.com/YosysHQ/yosys/blob/master/frontends/ast/ast.cc#L279. So it isn't needed to be called here.

kamilrakoczy: Fixed.

RRozak (approved):

PR #258, opened 2022-03-01
Convert AST_DOT with AST_RANGE to AST_PREFIX

RRozak:
It fixes [AssignGeneratedWireInAlways](https://github.com/chipsalliance/UHDM-integration-tests/pull/684/files#diff-72f82c5467315af03a48001b5dcf955e8357920f252194fc9c66aca4575263d3).

### DIFF
--- a/uhdm-plugin/UhdmAst.cc
+++ b/uhdm-plugin/UhdmAst.cc
@@ -841,30 +841,43 @@ static void simplify_format_string(AST::AstNode *current_node)
 
 static void simplify(AST::AstNode *current_node, AST::AstNode *parent_node)
 {
+    if (current_node->type == static_cast<int>(AST::AST_DOT))
+        current_node->type = AST::AST_IDENTIFIER;
+
+    auto dot_it =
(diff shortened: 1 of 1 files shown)


uhdm-plugin/UhdmAst.cc
```diff
@@ -841,30 +841,43 @@ static void simplify_format_string(AST::AstNode *current_node)
 
 static void simplify(AST::AstNode *current_node, AST::AstNode *parent_node)
 {
+    if (current_node->type == static_cast<int>(AST::AST_DOT))
+        current_node->type = AST::AST_IDENTIFIER;
```
kamilrakoczy, line 845: Why we need to do this? Is this because you later simplify ``dot`` to get constant?

If it is the case, please remove this and create temporary ``AST_IDENTIFIER`` and simplify it instead.

kamilrakoczy (changes requested):

uhdm-plugin/UhdmAst.cc
RRozak: Done

PR #260, opened 2022-03-02
Emit error when encountering assert/assume properties

thirtythreeforty:
Yosys doesn't have good handling of properties yet (see YosysHQ/yosys#3223).  In the meantime, SymbiYosys with the UHDM plugin will happily prove this model "correct":

```systemverilog
module wrong(input clk);
    logic data;
    always_comb data <= 0;
    assert property( @(posedge clk) data );
endmodule
```

The `assert property` is silently dropped from the UHDM parse tree when it's converted to AST nodes, and the user's design has a bug but passes formal.  (Changing it to a clocked immediate assertion allows the solver to find a counterexample.)

At least, emit an error for this case.  Real support for SVA assertions can hopefully be added later.


thirtythreeforty:
Also, @RRozak and @kamilrakoczy: I am unreasonably excited about all this work - the prospect of having open source, industry-grade formal verification tooling is awesome. Thank you!

PR #268, opened 2022-03-14
Removed const IdString class members

mkurc-ant:
This PR removes `const RTLIL::IdString` class members which caused Yosys to crash when the plugin code is built into Yosys binary.


mkurc-ant:
@coolbreeze413 Please take a look.

PR #270, opened 2022-03-16
Add 'read_systemverilog' command

tgorochowik:
This works as an alias to 'read_verilog_with_uhdm'


rkapuscik (approved):
LGTM as a first step, but we really should provide a default options set for Surelog library to make this easier to use.

kamilrakoczy (approved):

PR #272, opened 2022-03-16
Force -parse flag for Surelog when reading Verilog directly

tgorochowik:
These parameters are set if `-parse` is passed to Surelog

### DIFF
--- a/uhdm-plugin/uhdmsurelogastfrontend.cc
+++ b/uhdm-plugin/uhdmsurelogastfrontend.cc
@@ -110,6 +110,12 @@ struct UhdmSurelogAstFrontend : public UhdmCommonFrontend {
         if (!success) {
             log_error("Error parsing Surelog arguments!\n");
         }
+        // Force -parse flag settings even if it wasn't specified
+        clp->setwritePpOutput(true);
+        clp->setParse(true);
+        clp->setCompile(true);
+        clp->setElaborate(true);
+
         SURELOG::scompiler *compiler = nullptr;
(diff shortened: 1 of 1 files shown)


uhdm-plugin/uhdmsurelogastfrontend.cc
```diff
@@ -110,6 +110,12 @@ struct UhdmSurelogAstFrontend : public UhdmCommonFrontend {
         if (!success) {
             log_error("Error parsing Surelog arguments!\n");
         }
+        // Force -parse flag settings even if it wasn't specified
+        clp->setwritePpOutput(true);
```
kamilrakoczy, line 114: I'm not sure about this option, according to Surelog it is:
```
Writes out Preprocessor output (all compilation units will generate files under slpp_all/ or slpp_unit/)
```
I think we don't want to enable this by default.

kamilrakoczy (changes requested):

uhdm-plugin/uhdmsurelogastfrontend.cc
kamilrakoczy: But maybe it is more question to Surelog, as it is also set there when we add ``-parse`` argument.

tgorochowik: Are you sure it's safe to skip this? These 4 options are set when we pass the `-parse` flag which we require anyway:
https://github.com/chipsalliance/Surelog/blob/5c2bac5b95fc1d08f8b1b44a19391030ef069a02/src/CommandLine/CommandLineParser.cpp#L948-L952

rkapuscik (approved):

PR #274, opened 2022-03-17
fix the generation of carry chain for size <=2

tpagarani:
Signed-off-by: Tarachand Pagarani <tpagarani@quicklogic.com>


mkurc-ant (approved):
Ok, LGTM

PR #277, opened 2022-03-21
A fix for ql-qlf plugin full_adder test

mkurc-ant:
This PR disables check for specific LUT types and counts for one of the `ql-qlf` tests targeting the `PP3` architecture. The synthesis result seems to be not fully deterministic and varies depending on the way Yosys is built (Conda package vs. local build). In any case the equivalence check result is always positive.


mkurc-ant:
@tpagarani Please take a look.

acomodi (approved):
LGTM

PR #278, opened 2022-03-21
ql-qlf: qlf_k6n10f: bram: modify matching rules

lpawelcz:
This PR fixes a bug that is not allowing yosys to infer RAM in qlf_k6n10f device that has port width greater than 36 bits.
Removing the matching rules for `dbits` allows yosys to instantiate multiple RAM cells 'in parallel' allowing greater data port widths than the width of a single RAM cell for given device.
I also lowered the minimal required efficiency parameter for TDP variant and removed this rule entirely for SDP variant to allow inferring really small RAMs that will occupy only a small part of the actual RAM cell in hardware. 

### DIFF
--- a/ql-qlf-plugin/qlf_k6n10f/brams.txt
+++ b/ql-qlf-plugin/qlf_k6n10f/brams.txt
@@ -50,24 +50,22 @@ endbram
 
 
 match $__QLF_FACTOR_BRAM36_TDP
-  max dbits 36
-  max abits 15
   min wports 1
   max wports 2
   min rports 1
   max rports 2
-  min efficiency 2
+  min efficiency 1
+  min bits 128
   shuffle_enable B
   make_transp
   or_next_if_better
 endmatch
 
 match $__QLF_FACTOR_BRAM36_SDP
-  max dbits 36
-  max abits 15
   max wports 1
   max rports 1
-  min efficiency 2
+  min efficiency 1
+  min bits 128
   shuffle_enable B
   make_transp
 endmatch


lpawelcz:
@tpagarani Please, have a look at this PR

tpagarani (commented):
@lpawelcz could you please confirm the number of BRAMs being inferred for the testcase where I reported the issue?

lpawelcz:
Sure, it is 153 `TDP36K` cells

ql-qlf-plugin/qlf_k6n10f/brams.txt
```diff
@@ -50,24 +50,19 @@ endbram
...
   make_transp
   or_next_if_better
 endmatch
 
 match $__QLF_FACTOR_BRAM36_SDP
-  max dbits 36
-  max abits 15
   max wports 1
   max rports 1
-  min efficiency 2
```
tpagarani, line 70: What's the reason to remove the min efficiency for SDP mode altogether?

lpawelcz:
However, I have some problems with the synthesis process for that design. It freezes at techmap

ql-qlf-plugin/qlf_k6n10f/brams.txt
lpawelcz: Some of the really small RAMs have efficiency parameter close to 0 (the majority of TDP36K remain unused) and yosys rounds it to 0. If expected behavior is to infer all RAMs in design and map those to TDP36K blocks we need to remove this rule to allow inferring those small RAMs

lpawelcz: OK, I will adjust the rules to skip RAMs below 128bits.

tpagarani:
please go ahead and merge

PR #282, opened 2022-03-29
Set initial state of all k6n10f DSP registers to zero

mkurc-ant:
Besides that I have replaced all tabs with spaces in `cells_sim.v` hence the large diff.


mkurc-ant:

rakeshm75:
@mkurc-ant, the changes looks fine, its good to merge.

PR #291, opened 2022-04-06
Systemverilog: visit_object only when report or debug flag is preset

kamilrakoczy:
This PR fixes unnecessary call to visit_object when no ``debug`` flag or ``report`` flag is preset.
It also fixes several minor issues:

- limit arguments passed to Surelog only to arguments that were not handled by plugin
- makes sure that Surelog compilation succeed before trying to parse AST
- fixes help message for ``read_systemverilog`` command
- removes info about unhandled ``-process`` parameter
- adds support for ``-debug`` and ``-report`` arguments to ``read_systemverilog``/``read_verilog_with_uhdm`` frontends.


systemverilog-plugin/uhdmcommonfrontend.cc
```diff
@@ -78,8 +78,9 @@ void UhdmCommonFrontend::execute(std::istream *&f, std::string filename, std::ve
     bool dump_vlog2 = false;
     bool no_dump_ptr = false;
     bool dump_rtlil = false;
+    std::vector<std::string> unhandled_args;
 
-    for (size_t i = 1; i < args.size(); i++) {
```
rkapuscik, line 82: What is the reason for changing that? Was the first argument skipped? It is `1` in Verilog frontend, so I assume `args[0]` is the command name?

rkapuscik (changes requested):

systemverilog-plugin/uhdmcommonfrontend.cc
kamilrakoczy: Yes, ``args[0]`` is command name.
Surelog ``parseCommandLine`` expects that first argument is program name. Up to now, we were always passing whole args, so for ``read_systemverilog -debug test.v`` we were passing ``[read_systemverilog, -debug, test.v]`` to Surelog (but we were skipping ``read_systemverilog`` in our internal plugin loop. After this change, we pass ``[read_systemverilog, test.v]`` (as we already processed ``-debug`` flag).
We could add anything to the ``args[0]`` for Surelog, but I think it is better just to pass command name, as we are skipping it in our internal loop.

rkapuscik: Sure, passing the command in this case seems OK. Just add a comment explaining this, just in case something changes in `parseCommandLine` and we run into weird errors down the way.

kamilrakoczy: Ok, true, added.

rkapuscik (approved):

PR #306, opened 2022-04-22
Added support for overriding installation path via DESTDIR

mkurc-ant:
This PR adds support for the `DESTDIR` Makefile variable which allows to override the default plugin & data installation path provided by `yosys-config`.

### DIFF
--- a/Makefile_plugin.common
+++ b/Makefile_plugin.common
@@ -42,21 +42,29 @@
 YOSYS_PATH ?= $(realpath $(dir $(shell which yosys))/..)
 
 # Find yosys-config, throw an error if not found
-YOSYS_CONFIG ?= $(YOSYS_PATH)/bin/yosys-config
+YOSYS_CONFIG = $(YOSYS_PATH)/bin/yosys-config
 ifeq (,$(wildcard $(YOSYS_CONFIG)))
-$(error "Didn't find 'yosys-config' under '$(YOSYS_PATH)'")
+  $(error "Didn't find 'yosys-config' under '$(YOSYS_PATH)'")
 endif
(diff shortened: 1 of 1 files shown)


acomodi (approved):
LGTM

Makefile_plugin.common
```diff
@@ -42,22 +42,30 @@
...
-LDFLAGS ?= $(shell $(YOSYS_CONFIG) --ldflags)
-LDLIBS ?= $(shell $(YOSYS_CONFIG) --ldlibs)
-PLUGINS_DIR ?= $(shell $(YOSYS_CONFIG) --datdir)/plugins
-DATA_DIR ?= $(shell $(YOSYS_CONFIG) --datdir)
+CXXFLAGS = $(shell $(YOSYS_CONFIG) --cxxflags) #-DSDC_DEBUG
+LDFLAGS = $(shell $(YOSYS_CONFIG) --ldflags)
+LDLIBS = $(shell $(YOSYS_CONFIG) --ldlibs)
 EXTRA_FLAGS ?=
 
+ifdef DESTDIR
```
FFY00, line 56: This is unnecessary, the default value is empty, you can just prepend it where needed. Also, it's better hygiene to just put it in the commands that write to the filesystem, instead of changing the value of `PLUGINS_DIR`.

mkurc-ant: I don't think that I can simply prepend `DESTDIR` to the path provided by `yosys-config --destdir` command. The latter is an absolute path. I have to change `DATA_DIR` because it is referenced in plugin-specific Makefiles to know where to put additional data files.

PR #307, opened 2022-04-22
systemverilog: Add warnings to post inc/dec operations

tgorochowik:
It's okay to do this but we should warn as this may cause bugs in some cases.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -2575,7 +2575,11 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
         case vpiPowerOp:
             current_node->type = AST::AST_POW;
             break;
-        case vpiPostIncOp: // TODO: Make this an actual post-increment op (currently it's a pre-increment)
+        case vpiPostIncOp: {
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2570,7 +2570,13 @@ void UhdmAst::process_operation()
         case vpiPowerOp:
             current_node->type = AST::AST_POW;
             break;
-        case vpiPostIncOp: // TODO: Make this an actual post-increment op (currently it's a pre-increment)
+        case vpiPostIncOp: {
+            // TODO: Make this an actual post-increment op (currently it's a pre-increment)
+            const uhdm_handle *const handle = (const uhdm_handle *)obj_h;
+            const UHDM::BaseClass *const object = (const UHDM::BaseClass *)handle->object;
```
rkapuscik, line 2576: If we want `object` here for warning details, let's just pass it from `process_object` (like `process_event_control()`) and avoid code duplicaton.

rkapuscik (changes requested):

rkapuscik (approved):

PR #308, opened 2022-04-22
systemverilog: add dedicated wildcard operator errors

tgorochowik:
This should make it easier for users to understand the issues they're getting, for example here: https://github.com/antmicro/yosys-uhdm-plugin-integration/issues/679

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -2437,6 +2437,13 @@ void UhdmAst::process_operation()
     case vpiAssignmentPatternOp:
         process_assignment_pattern_op();
         break;
+    case vpiWildEqOp:
+    case vpiWildNeqOp: {
+        const uhdm_handle *const handle = (const uhdm_handle *)obj_h;
+        const UHDM::BaseClass *const object = (const UHDM::BaseClass *)handle->object;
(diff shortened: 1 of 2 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2437,6 +2437,13 @@ void UhdmAst::process_operation()
     case vpiAssignmentPatternOp:
         process_assignment_pattern_op();
         break;
+    case vpiWildEqOp:
+    case vpiWildNeqOp: {
+        const uhdm_handle *const handle = (const uhdm_handle *)obj_h;
+        const UHDM::BaseClass *const object = (const UHDM::BaseClass *)handle->object;
```
rkapuscik, line 2443: Same as https://github.com/chipsalliance/yosys-f4pga-plugins/pull/307#discussion_r857478705  - let's pass `object` here.

rkapuscik (changes requested):

systemverilog-plugin/UhdmAst.cc
tgorochowik: Fixed here, let's merge this first and then rebase #307

rkapuscik (approved):

PR #305, opened 2022-04-20
Systemverilog: adapt to current Surelog

rkapuscik:
* Update `import_typespec` case
* Handle shortint and time typespecs
* Fix packed ranges access in nets

PR #309, opened 2022-04-27
Use new Surelog/ include location.

hzeller:
Also: use instead of a comment, use [[fallthrough]] to let the
compiler know our intent (and squash warnings due to that).


hzeller:
(note, tests fail due to #305 not yet merged)

hzeller:
Can't assign reviewer, so here manually CC'ing you @kamilrakoczy

tgorochowik (approved):
LGTM once the CI is green, I'll also look into adding you here so you can assign people

hzeller:
I think the green-ness will happen after #305 is merged

kamilrakoczy (approved):

hzeller:
Alright, re-based and all tests pass.
I don't have permission to merge, so one of you needs to do that.

rkapuscik:
Thank you!

PR #323, opened 2022-05-18
Makefile: fix DESTDIR handling

Xiretza:
DESTDIR is an *additional* prefix to the installation path, not an *alternative*: https://www.gnu.org/prep/standards/html_node/DESTDIR.html


umarcor (approved):
LGTM!

tgorochowik (approved):

PR #330, opened 2022-05-20
Disable benign warnings.

hzeller:
In clang++ and g++-12, some warnings are triggered that are benign,
but would stop compilation as we compile with -Werror.

(TODO: mayabe we only enable -Werror in the CI, but are more
lenient otherwise to not have users 'discover' such things).


mithro (approved):
We should probably create GitHub issues for the items you mentioned.

PR #344, opened 2022-06-02
Systemverilog: Add separate compilation flow

rkapuscik:
This PR adds support for separate compilation flow by reusing Yosys `-defer` flag. 

Multiple files can be parsed one by one with Surelog by calling `read_systemverilog -defer <file>` and elaborating them later with `read_systemverilog -link`.

Needs https://github.com/chipsalliance/Surelog/pull/2991 for the CI to pass.


rkapuscik:
Integration PR: https://github.com/antmicro/yosys-uhdm-plugin-integration/pull/879

systemverilog-plugin/uhdmcommonfrontend.cc
```diff
@@ -53,16 +53,17 @@ void UhdmCommonFrontend::print_read_options()
...
     log("        write a coverage report for the UHDM file\n");
     log("\n");
     log("    -defer\n");
     log("        only read the abstract syntax tree and defer actual compilation\n");
     log("        to a later 'hierarchy' command. Useful in cases where the default\n");
     log("        parameters of modules yield invalid or not synthesizable code.\n");
+    log("        Needs to be followed by read_systemverilog -link after reading\n");
+    log("        all files.\n");
+    log("    -link\n");
+    log("        performs actual elaboration of the files read with -defer\n");
```
kbieganski, line 66: Do I understand it correctly that linking is a subset of elaboration? If so, the description should say that it's just linking, not full elaboration.

kbieganski (changes requested):

systemverilog-plugin/uhdmcommonfrontend.cc
rkapuscik: Actually this command does both in Surelog: `Links and elaborates the separately compiled files` (from its readme). Granted, we use only partial elaboration, but I'd argue that's irrelevant for user here.

I'm open for rewording this help, though. Would `performs linking and elaboration` work better here?

kbieganski: I think so. Right now, from a user perspective, it might be confusing why this option is called `-link`, a completely new word in this context, instead of `-elab` or something.

As an aside, I'd argue the use of `link` for this operation in Surelog is imprecise, though certainly we want to keep the same name to prevent confusion.

kbieganski (approved):

PR #352, opened 2022-06-15
Arith-stats plugin

tcal-x:


tcal-x:
I need to figure out  how to run the right `clang-format` with the right options locally.

PR #357, opened 2022-06-28
SystemVerilog: Support `break` and `continue`

kbieganski:
The general idea is to do transformations like:

```systemverilog
for (i = 0; i < 100; i++) begin
    <statements>
    continue;
    <statements>
end
```
to
```systemverilog
logic $continue;
for (i = 0; i < 100; i++) begin
    $continue = 0;
    <statements>
    $continue = 1;
    if (!$continue) begin
        <statements>
    end
end
```
and
```systemverilog
for (i = 0; i < 100; i++) begin
    <statements>
    break;
    <statements>
end
```
to
```systemverilog
logic $continue;
logic $break;
$break = 0;
for (i = 0; i < 100; i++) begin
    if (!$break) begin
        $continue = 0;
        <statements>
        $continue = 1;
        if (!$continue) begin
            <statements>
        end
    end
end
```

`while` loops are handled differently in that instead of wrapping the body in an `if`, the break condition gets ANDed with the original loop condition.

Integration PR: https://github.com/antmicro/yosys-uhdm-plugin-integration/pull/866

Signed-off-by: Krzysztof Bieganski <kbieganski@antmicro.com>

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1151,6 +1151,125 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
     return nullptr;
 }
 
+void UhdmAst::transform_breaks_continues(AST::AstNode *loop, AST::AstNode *decl_block)
+{
+    AST::AstNode *break_wire = nullptr;
+    AST::AstNode *continue_wire = nullptr;
+    // Creates a 1-bit wire with the given name
+    const auto make_cond_var = [this](const std::string &var_name) {
+        auto cond_var =
+          make_ast_node(AST::AST_WIRE, {make_ast_node(AST::AST_RANGE, {AST::AstNode::mkconst_int(0, false), AST::AstNode::mkconst_int(0, false)}),
+                                        AST::AstNode::mkconst_int(0, false)});
+        cond_var->str = var_name;
+        cond_var->is_reg = true;
+        return cond_var;
+    };
+    // Creates a conditional like 'if (!casevar) block'
+    auto make_case = [this](AST::AstNode *block, const std::string &casevar_name) {
+        auto *case_node = make_ast_node(AST::AST_CASE);
+        auto *id = make_identifier(casevar_name);
+        case_node->children.push_back(id);
+        auto *constant = AST::AstNode::mkconst_int(0, false, 1);
+        auto *cond_node = make_ast_node(AST::AST_COND);
+        cond_node->children.push_back(constant);
+        cond_node->children.push_back(block);
+        case_node->children.push_back(cond_node);
+        return case_node;
+    };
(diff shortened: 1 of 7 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -1151,6 +1151,127 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
...
+    }
+    if (break_wire) {
+        auto *break_id = make_ast_node(AST::AST_IDENTIFIER);
+        break_id->str = break_wire->str;
+        // Reset $break before the loop
+        auto *break_assign = make_ast_node(AST::AST_ASSIGN_EQ, {break_id, AST::AstNode::mkconst_int(0, false)});
+        decl_block->children.insert(decl_block->children.begin(), break_assign);
+        decl_block->children.insert(decl_block->children.begin(), break_wire);
+        if (loop->type == AST::AST_REPEAT || loop->type == AST::AST_FOR) {
+            // Wrap loop body in 'if (!break) {...}'
```
rkapuscik, line 1257: This works, but I wonder if with `for` shouldn't we amend the loop condition like in `while`?

rkapuscik (changes requested):
Really nice implementation, minor remarks only. Also, I'd say this is big enough to warrant a test or two [for the plugin itself](https://github.com/chipsalliance/yosys-f4pga-plugins/tree/main/systemverilog-plugin/tests), apart from the integration.

systemverilog-plugin/UhdmAst.cc
rkapuscik: Right, conditions in `for` are quite limited. A comment here would be helpful then.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1151,6 +1151,127 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
...
+                if (!continue_wire)
+                    continue_wire = make_cond_var("$continue");
+                auto *continue_id = make_ast_node(AST::AST_IDENTIFIER);
+                continue_id->str = continue_wire->str;
+                block->children.push_back(make_ast_node(AST::AST_ASSIGN_EQ, {continue_id, AST::AstNode::mkconst_int(1, false)}));
+                if (type == AST::AST_BREAK) {
+                    if (!break_wire)
+                        break_wire = make_cond_var("$break");
+                    auto *break_id = make_ast_node(AST::AST_IDENTIFIER);
+                    break_id->str = break_wire->str;
```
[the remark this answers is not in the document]
kbieganski: Refactored it into a separate function. Only used it in places affected by this PR.

kbieganski:
I added a little test for this. Might not be sufficient, but I'm not sure how comprehensive these plugin tests should be.

PR #359, opened 2022-06-29
SystemVerilog: Fix children handling in shift operations

rkapuscik:
Use `AST_TO_UNSIGNED` for shift operations. This mimics the way Yosys creates the AST.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -2663,14 +2663,14 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
         case vpiLShiftOp: {
             current_node->type = AST::AST_SHIFT_LEFT;
             log_assert(current_node->children.size() == 2);
-            auto unsigned_node = new AST::AstNode(AST::AST_TO_UNSIGNED, current_node->children[1]->clone());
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2672,16 +2660,20 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
         case vpiBitXnorOp:
             current_node->type = AST::AST_BIT_XNOR;
             break;
-        case vpiLShiftOp:
+        case vpiLShiftOp: {
             current_node->type = AST::AST_SHIFT_LEFT;
             log_assert(current_node->children.size() == 2);
-            mark_as_unsigned(current_node->children[1], object);
+            auto unsigned_node = new AST::AstNode(AST::AST_TO_UNSIGNED, current_node->children[1]->clone());
+            current_node->children[1] = unsigned_node;
```
kamilrakoczy, line 2667: I think we don't need to clone ``current_node->children[1]`` (and the same applies to other cases) as it is replaced in the next line, but please confirm.

kamilrakoczy (changes requested):
I minor comment, otherwise looks good.

PR #361, opened 2022-07-01
Added back regular D flip-flops and latches for k6n10f

mkurc-ant:


mkurc-ant:
@tpagarani This PR re-adds regular DFFs and latches. I've checked in the architecture and I couldn't find `latchn` mode for IOB so I haven't added it. Shall I add it anyway?

rakeshm75:
@mkurc-ant, There is no need to add latch or latchn, basically we want to add dff and dffn to use the flops in the IO block. The block in the IO does not support the latch mode.

mkurc-ant:
@rakeshm75 Alright, I've removed the `latch` model.

rakeshm75:
@mkurc-ant, this looks fine. Is there another PR for adding a flag to differentiate the clb with and without sync reset/set in Yosys?

mkurc-ant:
@rakeshm75 I'm not aware of any different CLB types. I only know that `dff` and `dffn` are needed for IO tiles which do not have reset / set inputs. `dff` and `dffn` cells are not inferred hence the synthesis flow is not change. One has to insert them manually to make VPR pack them into IO tiles.

tpagarani (approved):
lgtm

PR #363, opened 2022-07-06
ql-qlf: k6n10f: Add QL_DSP3 DSP flavor

lpawelcz:
This PR adds second flavor of DSP cells - QL_DSP3. These cells use configuration parameters instead of ports (as it is in QL_DSP2) to pass configuration bits.
PR: 
* moves all DSP simulation modules to separate file `dsp_sim.v`
* modifies existing techmaps and adds others for QL_DSP3
* adjusts `ql-dsp-*` passes to support QL_DSP3
* adds flag `-use_dsp_cfg_params` to `synth_quicklogic` pass so that user can choose which DSP variant should be inferred
* updates tests, improves tcl scripts, adds new test cases
* it also introduces changes from https://github.com/chipsalliance/yosys-f4pga-plugins/pull/340


lpawelcz:
@tpagarani @rakeshm75 Please take a look

PR #369, opened 2022-07-19
Fix k6n10f DSP type inference

mkurc-ant:
This PR fixes the issue of incorrect DSP type inference and incorrect clock port removal. It adds a test for multiply+add mode type change.

Due to the `feedback_i` port controlling the feedback loop of the accumulator it is not possible to tell if the requested operation is MACC or MADD if the connected signal is not a constant.


mkurc-ant:

tpagarani (approved):

PR #373, opened 2022-07-29
[k6n10f] Preserve manually inserted DSPs

mkurc-ant:
This PR adds preservation of all manually inserted `QL_DSP2` cells - their types will never get changed.

The feature is implemented by annotating inferred DSP cells with an attribute which tells the pass that changes types that the change is allowed.


mkurc-ant:

tpagarani (approved):

PR #379, opened 2022-08-25
Fix QuickLogic asymmetric BRAM inference

mkurc-ant:
This PR fixes the inference mechanism by changing the way BRAM wires are looked up which does not trigger Yosys assertions in specific cases. Moreover, it also ensures that inferred BRAM cell names remain unique.


tpagarani:
lgtm

PR #381, opened 2022-08-30
[k6n10f] RAM functional modes for TDP36K

mkurc-ant:
This PR adds support for functional modes of TDP36K in non-split and split mode.

For each inferred `TDP36K` cell Yosys will now replace it with a cell type named `TDP36K_BRAM_WR_X<N>_RD_X<M>_nonsplit` or `TDP36K_BRAM_WR_X<N>_RD_X<M>_split` where `N` denotes configured write bus width and `M` read bus width respectively. Manually inserted BRAM cells will be left intact. The "write" port is the port A while the "read" port corresponds to the port "B".


tpagarani:
@rakeshm75  could you please validate this change through complete openfpga flow to make sure that it passes.

mkurc-ant:
@tpagarani @rakeshm75 I have added emitting functional mode cells for split BRAMs as well.

rakeshm75:
@mkurc-ant, I have run through the OpenFPGA flow and ran the simulations, Post layout simulation and Bitstream simulation and it works fine. But following are the observations/ modifications required:

1. The compilation of brams_sim.v file gives lots of errors like below:
# ** Error: D:/Tamar2/System_Engineering/AntMicro/QLSOFA/K6N10_Alpha/bram_sim/lib/brams_sim.v(807): (vlog-2892) Net type of 'A1ADDR' must be explicitly declared.
This is because the wire needs to be explicitly declared for input and outputs as `default_nettype none is used.

2. Secondly, in the Synthesis Verilog output file (*_output_verilog.v) the port-mapping vector signals are split:

EX:                                
              
  TDP36K_BRAM_WR_X9_RD_X9_nonsplit #(
                                 .MODE_BITS(81'h001402809240140140492)
                                ) _3_ (
                                 .\ADDR_A1_i[0] (_0_),
                                 .\ADDR_A1_i[10] (\ra[7] ),
                                 .\ADDR_A1_i[11] (\ra[8] ),
                                .\ADDR_A1_i[12] (\ra[9] ),
                                .\ADDR_A1_i[13] (\ra[10] ),
                                .\ADDR_A1_i[14] (\ra[11] ),
                                             .
                                             .
                                             .

This gives port map mismatch errors.

mkurc-ant:
@rakeshm75 I have updated the cell definitions so that all inputs/outputs have their types explicitly declared. This should solve the problem with multitude of errors.

As for split multi-bit ports please try specifying the `--gen_post_implementation_merged_netlist on` option in VPR (https://docs.verilogtorouting.org/en/latest/vpr/command_line_usage/#cmdoption-vpr-gen_post_implementation_merged_netlist)

If that won't solve the problem then we would probably need to work on modifying VPR to aggregate multi-bit ports of cell instances.

mkurc-ant:
Hi @rakeshm75 

We tested the flow on our side and we don't see BRAM ports being split into individual bits. It looks like the change to VPR responsible for aggregating single-bit ports to vectors is already present in the VPR used in OpenFPGA.

To verify that we used the Aurora2 toolchain SHA `40f9eab9b31ebf39b597e93de6d8bde87db0f043` which uses OpenFPGA SHA `2a43843bd7a7e73f2fdfd9563d34cdb3016d4051`. Could you provide SHAs of versions you used that gives the error?

tpagarani:
@mkurc-ant can we put the expansion of BRAM macros under a new flag in synth_quicklogic.cc and turn if off by default. We would like to control this feature with a flag. Once you do that this PR should be good to merge.

ql-qlf-plugin/synth_quicklogic.cc
```diff
@@ -368,6 +370,29 @@ struct SynthQuickLogicPass : public ScriptPass {
             if (family == "qlf_k6n10f") {
```
tpagarani, line 370: Please put this code under a new flag to make it controllable

tpagarani:
@mkurc-ant Can you merge this PR after putting the code under a flag as I requested

PR #383, opened 2022-09-25
Fix deprecated use of `which`

hzeller:
Instead of `which`, the suggested alternative
is `command -v`


hzeller:
Context: https://salsa.debian.org/debian/debianutils/-/blob/5aeca1e3b14c08b5ab92995e91efcc2b3806a639/debian/NEWS#L7-16

PR #387, opened 2022-10-11
Support signed ports

rkapuscik:
Correctly find and annotate type for signed ports.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3528,9 +3528,7 @@ void UhdmAst::process_logic_typespec()
     }
     visit_one_to_many({vpiRange}, obj_h, [&](AST::AstNode *node) { packed_ranges.push_back(node); });
     add_multirange_wire(current_node, packed_ranges, unpacked_ranges);
-    if (vpi_get(vpiSigned, obj_h) == true) {
-        current_node->is_signed = true;
-    }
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -3527,6 +3528,9 @@ void UhdmAst::process_logic_typespec()
     }
     visit_one_to_many({vpiRange}, obj_h, [&](AST::AstNode *node) { packed_ranges.push_back(node); });
     add_multirange_wire(current_node, packed_ranges, unpacked_ranges);
+    if (vpi_get(vpiSigned, obj_h) == true) {
+        current_node->is_signed = true;
+    }
```
kamilrakoczy, line 3533: Do we only want to set ``is_signed`` when it is ``true`` (what about the case, when ``current_node->is_signed`` is ``true`` and ``vpi_get(vpiSigned, obj_h)`` is ``false``)? Maybe something like this would be better:
```suggestion
    current_node->is_signed = vpi_get(vpiSigned, obj_h);
```

rkapuscik: `is_signed` is always initialized to `false`, so we shouldn't run into this case. I agree that always setting the status is clearer though, as we check it anyway. Changed.

kamilrakoczy (approved):

PR #388, opened 2022-10-13
ql-qlf: k6n10f: ql_dsp_simd: change naming scheme

lpawelcz:
RTLIL::Cell 'name' field contains path to source verilog file. Assigning two of those as the name of new DSP cell will likely cause generation of very long module names, especially when the path argument to 'read_verilog' is a long absolute path.
Long module names will propagate to long net names through 'autoname' pass and cause readability issues in post-synthesis verilog. This change prevents that.

Signed-off-by: Pawel Czarnecki <pczarnecki@antmicro.com>


lpawelcz:
Hi @tpagarani, please have a look at this.

tpagarani (approved):
lgtm

PR #391, opened 2022-10-26
handle int constants wider than 32 bits

wsipak:
The problem is described here:
https://github.com/antmicro/yosys-systemverilog/issues/1043

We've observed that even though this integer is mishandled:
```
assign a = 7698294523898761276;
```
Integers written as based literal constants with the same value are processed properly:
```
assign a = 'sd7698294523898761276;
```

The solution makes use of the already existing code for the latter case, and now integers that exceed 32 bits are treated as strings, just like based literal constants.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1,6 +1,7 @@
 #include <algorithm>
 #include <cstring>
 #include <functional>
+#include <limits>
 #include <regex>
 #include <string>
 #include <vector>
@@ -1094,7 +1095,22 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
         // Surelog reports constant integers as a unsigned, but by default int is signed
         // so we are treating here UInt in the same way as if they would be Int
         case vpiUIntVal:
+            if (val.value.uint > std::numeric_limits<std::uint32_t>::max()) {
+                // an integer is by default signed, so use 'sd despite the variant vpiUIntVal
+                strValType = "'sd";
+                string str_value = std::to_string(val.value.uint);
+                val.value.str = strdup(str_value.c_str());
+                break;
+            }
+            [[fallthrough]];
         case vpiIntVal: {
+            if (val.value.integer > std::numeric_limits<std::int32_t>::max()) {
+                strValType = "'sd";
+                string str_value = std::to_string(val.value.integer);
+                val.value.str = strdup(str_value.c_str());
+                break;
+            }
+
             int size = -1;
             bool is_signed = false;
             // Surelog sometimes report size as part of vpiTypespec (e.g. int_typespec)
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -1088,7 +1089,22 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
...
+            }
+            [[fallthrough]];
         case vpiIntVal: {
+            if (val.value.integer > std::numeric_limits<std::int32_t>::max()) {
+                strValType = "'sd";
+                string str_value = std::to_string(val.value.integer);
+                val.value.str = strdup(str_value.c_str());
+                break;
+            }
+
```
kamilrakoczy, line 1107: Later we are always using ``val.value.integer`` (https://github.com/chipsalliance/yosys-f4pga-plugins/blob/main/systemverilog-plugin/UhdmAst.cc#L1113) for both ``vpiUIntVal`` and ``vpiIntVal``. Maybe we can simplify this to the single case, or we should fix usage of ``val.value.integer`` in case of ``vpiUIntVal``

wsipak: Alright, I've changed this so that now the proper variant from `val.value` is used in the call to `mkconst_int`.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1088,7 +1089,22 @@ AST::AstNode *UhdmAst::process_value(vpiHandle obj_h)
...
+                strValType = "'sd";
+                string str_value = std::to_string(val.value.uint);
+                val.value.str = strdup(str_value.c_str());
+                break;
+            }
+            [[fallthrough]];
         case vpiIntVal: {
+            if (val.value.integer > std::numeric_limits<std::int32_t>::max()) {
+                strValType = "'sd";
+                string str_value = std::to_string(val.value.integer);
```
rkapuscik, line 1103: I wonder if this case can be simplified further. Will parsing the strings return incorrect values for smaller constants? Maybe we just need to set the size in that case and then parse them from strings?

rkapuscik (changes requested):
Please also add tests for those constants. Some ideas:
* Test 32, 64 bit constants
* Test constants one bit larger and smaller than those 
* Test an even bigger constant with some odd size

systemverilog-plugin/UhdmAst.cc
wsipak: Parsing decimals as strings can also be used for numbers shorter than 32 bits with success. However, I left this case deliberately to be handled by the older flow which considers Typespec before creating the constant Node. In case when we use strings, we don't actually set any size. This is done by the function which parses the string. In this case, it's `VERILOG_FRONTEND::const2ast`.

I've added new test cases here:
https://github.com/chipsalliance/UHDM-integration-tests/pull/696

wsipak: @rkapuscik  ping, could you take a look at this again? I think it's ready, although I removed the test case exceeding 64 bits, as this PR only covers <32;64> bits and I've run tests without bumping other submodules.
Tests were successful: https://github.com/antmicro/yosys-systemverilog/pull/1196
To my knowledge, there's nothing else I would be expected to change.

rkapuscik: OK, I think it's acceptable for this limited scope. LGTM.

rkapuscik (approved):

PR #392, opened 2022-10-27
systemverilog: Support op type on assignment

mandrys:
Compound assignments used to be interpreted as simple assignments, like it is described in this issue: antmicro/yosys-systemverilog#1050. This was caused by the fact, that _vpiOpType_ of _assignment_ node in UHDM tree was not taken into account.
This PR introduces operation type check in the assignment node and expanding it to include operation sub-nodes in cases where it is needed.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -2265,22 +2265,20 @@ void UhdmAst::process_cont_assign()
 void UhdmAst::process_assignment(const UHDM::BaseClass *object)
 {
     auto type = vpi_get(vpiBlocking, obj_h) == 1 ? AST::AST_ASSIGN_EQ : AST::AST_ASSIGN_LE;
-    bool simple_assign = true;
     bool shift_unsigned = false;
     int op_type = vpi_get(vpiOpType, obj_h);
     AST::AstNodeType node_type;
     current_node = make_ast_node(type);
 
-    if (op_type && op_type != vpiAssignmentOp) {
-        simple_assign = false;
-        visit_one_to_one({vpiLhs}, obj_h, [&](AST::AstNode *node) {
-            if (node) {
-                if (node->type == AST::AST_PARAMETER || node->type == AST::AST_LOCALPARAM) {
-                    node->type = AST::AST_IDENTIFIER;
-                }
-                current_node->children.push_back(node);
+    visit_one_to_one({vpiLhs, vpiRhs}, obj_h, [&](AST::AstNode *node) {
+        if (node) {
+            if (node->type == AST::AST_PARAMETER || node->type == AST::AST_LOCALPARAM) {
+                node->type = AST::AST_IDENTIFIER;
             }
-        });
+            current_node->children.push_back(node);
+        }
+    });
+    if (op_type && op_type != vpiAssignmentOp) {
         switch (op_type) {
         case vpiSubOp:
             node_type = AST::AST_SUB;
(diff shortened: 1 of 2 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2264,16 +2264,93 @@ void UhdmAst::process_cont_assign()
 
 void UhdmAst::process_assignment()
 {
+    const uhdm_handle *const handle = (const uhdm_handle *)obj_h;
+    const UHDM::BaseClass *const object = (const UHDM::BaseClass *)handle->object;
```
rkapuscik, line 2268: It's already done in `process_object`, we can just pass it to `process_assignment` like in other functions.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2264,16 +2264,93 @@ void UhdmAst::process_cont_assign()
...
     auto type = vpi_get(vpiBlocking, obj_h) == 1 ? AST::AST_ASSIGN_EQ : AST::AST_ASSIGN_LE;
+    bool simple_assign = true;
+    bool shift_unsigned = false;
+    int op_type = vpi_get(vpiOpType, obj_h);
+    AST::AstNodeType node_type;
     current_node = make_ast_node(type);
+
+    if (op_type && op_type != 82) {
```
rkapuscik, line 2276: Use the define instead of magic number.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2264,16 +2264,93 @@ void UhdmAst::process_cont_assign()
...
+            delete current_node;
+            current_node = nullptr;
+            report_error("%s:%d: Encountered unhandled compound assignment with operation type %d\n", object->VpiFile().c_str(), object->VpiLineNo(),
+                         op_type);
+            return;
+        }
+        current_node->children.push_back(make_ast_node(node_type));
+    }
+
     visit_one_to_one({vpiLhs, vpiRhs}, obj_h, [&](AST::AstNode *node) {
```
rkapuscik, line 2337: We don't need to visit the LHS again, we can just copy the node obtained earlier.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2264,16 +2264,93 @@ void UhdmAst::process_cont_assign()
...
         if (node) {
             if (node->type == AST::AST_PARAMETER || node->type == AST::AST_LOCALPARAM) {
                 node->type = AST::AST_IDENTIFIER;
             }
-            current_node->children.push_back(node);
+            if (!simple_assign) {
```
rkapuscik, line 2342: In general we can always visit LHS and RHS first, then assemble the RHS depending on operation type (for simple assignment passing it directly, for others assembling the operation).

rkapuscik (changes requested):

systemverilog-plugin/UhdmAst.cc
mandrys: Done. I added my own define, as there are only defines for values 1-43 for _vpiOpType_, but in case of simple assignment we always have 82.

mandrys:
I applied requested changes in the new commits which I will squash after the changes are approved.

systemverilog-plugin/UhdmAst.cc
mandrys: OK, modified the code to use _clone()_ method instead of visiting LHS for the twice.

systemverilog-plugin/UhdmAst.cc
mandrys: In fact, there is a define for _vpiAssignmentOp_ (here: https://github.com/chipsalliance/UHDM/blob/019f05e14a018de89b8ae2f409cb1391b1f5fe91/include/sv_vpi_user.h#L390), so changed my previous modification to use this one.

systemverilog-plugin/UhdmAst.cc
rkapuscik: That looks much better, thanks!

rkapuscik (approved):

PR #395, opened 2022-11-04
ql-qlf: Split BRAM into cells with equal types only

robertszczepanski:
Currently BRAM Split pass reports an error when it encounters SDP and TDP cells next to each other. This PR ensures that SDP and TDP cells are not mixed up and it maps every pair of the same types into one TDP36K BRAM if possible.


mkurc-ant (approved):
LGTM

PR #397, opened 2022-11-15
Minimize diffs between frontends

rkapuscik:
This PR adds some features that should make comparing diffs between the plugin and the default frontend easier:
* remove additional ranges attributes that were already converted
* retrieve Verilog line and column information from UHDM

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1277,8 +1277,17 @@ AST::AstNode *UhdmAst::make_ast_node(AST::AstNodeType type, std::vector<AST::Ast
     if (auto filename = vpi_get_str(vpiFile, obj_h)) {
         node->filename = filename;
     }
-    if (unsigned int line = vpi_get(vpiLineNo, obj_h)) {
-        node->location.first_line = node->location.last_line = line;
+    if (unsigned int first_line = vpi_get(vpiStartLine, obj_h)) {
+        node->location.first_line = first_line;
+    }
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -1277,8 +1277,17 @@ AST::AstNode *UhdmAst::make_ast_node(AST::AstNodeType type, std::vector<AST::Ast
...
+        node->location.first_line = first_line;
+    }
+    if (unsigned int last_line = vpi_get(vpiEndLineNo, obj_h)) {
+        node->location.last_line = last_line;
+    }
+    if (unsigned int first_col = vpi_get(vpiColumnNo, obj_h)) {
+        node->location.first_column = first_col;
+    }
+    if (unsigned int last_col = vpi_get(vpiEndColumnNo, obj_h)) {
+        node->location.last_column = last_col;
```
mglb, line 1290: What if there is non-zero `first_line` (1st `if` is true), but the `last_line` is zero (2nd `if` is false)?
Shouldn't there be a fallback like `node->location.first_line = node->location.last_line = first_line;` in the first `if`?

rkapuscik: They are initialized to 0, but the fallback makes sense for the last line/column. Added.
If the first line/column is unavailable, they will just default to 0;

mglb (approved):
LGTM, also all green in yosys-systemverilog https://github.com/antmicro/yosys-systemverilog/actions/runs/3481136178

PR #408, opened 2022-11-24
process task call nodes

wsipak:
Previously, the plugin would just create a node of type `AST_TCLL`, while Yosys also processes the parameters of the task, and creates `AST_IDENTIFIER` nodes accordingly, inside the `AST_TCLL`.
With this change, the plugin processes the parameters, too.

This solves the issue described here: https://github.com/antmicro/yosys-systemverilog/issues/1137

See tests here:
https://github.com/antmicro/yosys-systemverilog/pull/1293

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3601,9 +3601,9 @@ void UhdmAst::process_sys_func_call()
         current_node->str = current_node->str.substr(1);
 }
 
-void UhdmAst::process_func_call()
+void UhdmAst::process_tf_call(AST::AstNodeType type)
 {
-    current_node = make_ast_node(AST::AST_FCALL);
+    current_node = make_ast_node(type);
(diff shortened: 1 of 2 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -3615,6 +3615,20 @@ void UhdmAst::process_func_call()
     });
 }
 
+void UhdmAst::process_task_call()
+{
+    current_node = make_ast_node(AST::AST_TCALL);
```
rkapuscik, line 3620: It seems that apart from this type, the processing is identical to function calls (and it also handles `localparam`s). If that's true, it would make sense to handle it in a common function, like `process_tf_call`.

rkapuscik (changes requested):

systemverilog-plugin/UhdmAst.cc
wsipak: Thank you for your feedback! Alright, I've merged the two functions into `process_tf_call`.

rkapuscik (approved):

PR #418, opened 2022-12-09
systemverilog: Fix memory inference

robertszczepanski:
Fixes https://github.com/antmicro/yosys-systemverilog/issues/1204.

Memory is now inferred correctly and produced `_gate.v` is not different for Surelog and clean Yosys.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -340,7 +340,7 @@ static void resolve_wiretype(AST::AstNode *wire_node)
     log_assert(AST_INTERNAL::current_scope.count(wiretype_node->str));
     wiretype_ast = AST_INTERNAL::current_scope[wiretype_node->str];
     // we need to setup current top ast as this simplify
-    // needs to have access to all already definied ids
+    // needs to have access to all already defined ids
     while (wire_node->simplify(true, false, false, 1, -1, false, false)) {
     }
     if (wiretype_ast->children[0]->type == AST::AST_STRUCT && wire_node->type == AST::AST_WIRE) {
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -414,11 +414,29 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
...
+            if (wire_node->type == AST::AST_LOCALPARAM)
+                return true;
+            if ((wire_node->is_input || wire_node->is_output) && (packed_ranges.size() > 0 || unpacked_ranges.size() > 0))
+                return true;
+            if (wire_node->attributes.count(UhdmAst::force_convert()) && wire_node->attributes[UhdmAst::force_convert()]->integer == 1)
+                return true;
+            return false;
+        } else {
+            return false;
+        }
```
mglb, line 438: No logic changes, just one nesting level less:
```suggestion
        if (wire_node->type == AST::AST_MEMORY)
            return false;
        if (packed_ranges.size() > 1)
            return true;
        if (unpacked_ranges.size() > 1)
            return true;
        if (wire_node->attributes.count(ID::wiretype))
            return true;
        if (wire_node->type == AST::AST_PARAMETER)
            return true;
        if (wire_node->type == AST::AST_LOCALPARAM)
            return true;
        if ((wire_node->is_input || wire_node->is_output) && (packed_ranges.size() > 0 || unpacked_ranges.size() > 0))
            return true;
        if (wire_node->attributes.count(UhdmAst::force_convert()) && wire_node->attributes[UhdmAst::force_convert()]->integer == 1)
            return true;
        return false;
```

mglb (approved):
One small and optional simplification suggestion. LGTM overall.

systemverilog-plugin/UhdmAst.cc
robertszczepanski: Thanks for the suggestion, I applied it.

PR #419, opened 2022-12-14
systemverilog: Fix multiple function return statements

robertszczepanski:
Currently all return statements are included in AST tree which results in incorrect multiple assign with different values like:
```verilog
module top(o);
  output o;
  /** AST_FUNCTION **/
  assign o = 1'b 0;
  assign o = 1'b 1;
endmodule
```
This PR fixes https://github.com/antmicro/yosys-systemverilog/issues/1243.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -73,6 +73,12 @@ enum AstNodeTypeExtended {
     return id;
 }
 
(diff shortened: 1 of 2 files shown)


robertszczepanski:
This issue is resolved by some commits since December when I started working on it. Test passes without these changes and generated gate level looks good, so I'm closing this PR.

PR #420, opened 2022-12-14
systemverilog-plugin: use dedicated namespace and move code from const2ast.cc to third_party directory.

mglb:
The code in `UhdmAstUpstream.cc` has been moved to `third_party/yosys/const2ast.cc`. It was mostly a copy of const2ast.cc from Yosys sources. Let's not mix external and our own code when not required and
keep thirdparty code in its own directory with clear source information.

Remaining code from UhdmAstUpstream.{h,cc} (enum and mkconst_real) has been moved to UhdmAst.cc. The enum has been enclosed in additional namespace to indicate that the values are extensions.

All systemverilog-plugin code has been enclosed in `systemverilog_plugin` namespace. This fixes issue with static linking mentioned in https://github.com/chipsalliance/yosys-f4pga-plugins/commit/f95e1b244cfb54b67c3c7723e35d137359b4a132#commitcomment-92399415 and will prevent symbol name collision in the future.

yosys-systemverilog CI run: 
This also fixes issue with static linking mentioned in https://github.com/chipsalliance/yosys-f4pga-plugins/commit/f95e1b244cfb54b67c3c7723e35d137359b4a132#commitcomment-92399415

### DIFF
--- a/systemverilog-plugin/third_party/yosys/README
+++ b/systemverilog-plugin/third_party/yosys/README
@@ -1,9 +1,16 @@
 Files in this directory were copied from Yosys sources and slightly adapted.
 Original sources and their license available at https://github.com/YosysHQ/yosys.
 
-Copied files and their sources:
+Copied files, their sources, changes & notes:
 
-- const2ast.cc: yosys/frontends/verilog/const2ast.cc
+- const2ast.cc: yosys/frontends/verilog/const2ast.cc (rev. 72787f5)
+  - The file is a part of Yosys Verilog frontend, which is not publicly exposed
+    by Yosys. Copy has been made to avoid relying on internal details.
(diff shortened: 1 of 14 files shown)


systemverilog-plugin/third_party/yosys/README
```diff
@@ -0,0 +1,10 @@
+Files in this directory were copied from Yosys sources and slightly adapted.
```
hzeller, line 1: Might be worth-while mentioning what adaption has been done, and why we can't link against the yosys implementation.

mglb: Added reason, changes, and exact source revision

rkapuscik (approved):

PR #421, opened 2022-12-15
Parmys plugin

poname:
This PR provides _Parmys_ (Partial Mapper for Yosys) plugin which performs intelligent partial mapping (inference, binding, and hard/soft logic trade-offs) from [Odin-II](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/ODIN_II).


poname:
@kgugala once the tests are passed, this PR is ready to be reviewed/merged.

parmys-plugin/parmys.cc
```diff
@@ -0,0 +1,1175 @@
...
+
+struct ParMYSPass : public Pass {
+
+    static void hook_up_nets(netlist_t *odin_netlist, Hashtable *output_nets_hash)
+    {
+        nnode_t **node_sets[] = {odin_netlist->internal_nodes, odin_netlist->ff_nodes, odin_netlist->top_output_nodes};
+        int counts[] = {odin_netlist->num_internal_nodes, odin_netlist->num_ff_nodes, odin_netlist->num_top_output_nodes};
+        int num_sets = 3;
+
+        int i;
```
kgugala, line 78: why not declare this in the loop definition?

parmys-plugin/parmys.cc
```diff
@@ -0,0 +1,1175 @@
...
+    {
+        int j;
+        for (j = 0; j < node->num_input_pins; j++) {
+            npin_t *input_pin = node->input_pins[j];
+
+            nnet_t *output_net = (nnet_t *)output_nets_hash->get(input_pin->name);
+
+            if (!output_net)
+                log_error("Error: Could not hook up the pin %s: not available, related node: %s.", input_pin->name, node->name);
+            add_fanout_pin_to_net(output_net, input_pin);
```
kgugala, line 98: this will be called even if `output_net` is NULL (which you check above). Shouldn't this function simply return on the error (and maybe return some error value so caller can react(?

parmys-plugin/include/odin_types.h
```diff
@@ -0,0 +1,671 @@
+#ifndef ODIN_TYPES_H
```
kgugala, line 1: there seems to be a few files copied from VPR project. I don't think this is a good idea to do this. You should rather add a dependency of VPR and use it. See the `systemverilog` plugin (https://github.com/chipsalliance/yosys-f4pga-plugins/blob/main/systemverilog-plugin/Makefile#L35).

kgugala (changes requested):
I've added a few initial comments. This PR is pretty big and I think the biggest issue is that it seems to copy code from other projects instead of adding them as a third_party dependency.

This will need to be solved before we can land this PR

parmys-plugin/include/BlockMemories.hpp
```diff
@@ -0,0 +1,100 @@
+/**
+ * Copyright (c) 2021 Seyed Alireza Damghani (sdamghann@gmail.com)
+ *
+ * Permission is hereby granted, free of charge, to any person
```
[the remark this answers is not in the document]
poname: The licensing issue is resolved.
(adding @kennethkent as the project owner)

parmys-plugin/parmys.cc
poname: refined the `for` loop syntax within the whole code base of the plugin.

parmys-plugin/parmys.cc
poname: AFAIK the `log_error` function of Yosys throws an exception while logging the error. 
And here we expect the runtime to be interrupted if `output_net` is NULL.
You mean if `output_net` is NULL the `add_fanout_pin_to_net` is still called?

poname:
When using `litex-hub::vtr` package as VTR dependency, the ci gets the following error while performing linking to the vtr binaries:
```
  lto1: fatal error: bytecode stream in file ‘/home/runner/work/yosys-f4pga-plugins/yosys-f4pga-plugins/env/conda/envs/yosys-plugins/lib/libarchfpga.a’ generated with LTO version 12979.5 instead of the expected 8.1
  compilation terminated.
  lto-wrapper: fatal error: /usr/bin/g++ returned 1 exit status
  compilation terminated.
  /usr/bin/ld: error: lto-wrapper failed
  collect2: error: ld returned 1 exit status
```

@kgugala do you have any idea what this would be?

I also checked with Tim:
#https://github.com/hdl/conda-eda/pull/272#issuecomment-1367641019

parmys-plugin/include/odin_types.h
poname: this one is resolved.

PR #425, opened 2022-12-21
Don't visit vpiContAssign in uhdmallModules

mandrys:
Since `vpiContAssign` node is now present in _uhdmtopModules_ and all needed info is present there, we shouldn't visit it in _uhdmallModules_, otherwise we get duplicated AST_ASSIGN nodes as the result.


mandrys:
The plugin with these changes is built and tested here: https://github.com/antmicro/yosys-systemverilog/pull/1352

mglb (approved):

PR #426, opened 2022-12-22
Handle -formal

rkapuscik:
This adds support for the `-formal` option in the plugin, setting defines in the same way as Yosys parser and passing the option to Surelog to include verification constructs in UHDM.

### DIFF
--- a/systemverilog-plugin/uhdmcommonfrontend.cc
+++ b/systemverilog-plugin/uhdmcommonfrontend.cc
@@ -120,6 +120,10 @@ void UhdmCommonFrontend::execute(std::istream *&f, std::string filename, std::ve
             this->shared.link = true;
             // Surelog needs it in the command line to link correctly
             unhandled_args.push_back(args[i]);
+        } else if (args[i] == "-formal") {
+            this->shared.formal = true;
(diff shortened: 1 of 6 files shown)


systemverilog-plugin/uhdmcommonfrontend.cc
```diff
@@ -120,6 +120,10 @@ void UhdmCommonFrontend::execute(std::istream *&f, std::string filename, std::ve
             this->shared.link = true;
             // Surelog needs it in the command line to link correctly
             unhandled_args.push_back(args[i]);
+        } else if (args[i] == "-formal") {
```
mandrys, line 123: We could add this new switch to be printed out by _print_read_options_, [line 33](https://github.com/chipsalliance/yosys-f4pga-plugins/blob/6b61eb85b9042aabb4e32d57d37891fa846e8815/systemverilog-plugin/uhdmcommonfrontend.cc#L33).

rkapuscik: Added (and fixed some missing newlines).

mandrys (approved):

PR #428, opened 2022-12-27
submodules: Update make-env

tmichalak:
Signed-off-by: Tomasz Michalak <tmichalak@antmicro.com>


mithro:
@tmichalak - Anytime you pin something you should log a github issue which explains when the pin can be removed. Can you please do that?

tmichalak:
> @tmichalak - Anytime you pin something you should log a github issue which explains when the pin can be removed. Can you please do that?

Sure thing, done.

PR #433, opened 2023-01-09
Split Github Actions

apokusinski:
Fixes https://github.com/chipsalliance/yosys-f4pga-plugins/issues/432
The goal is to split apart GH Actions, so that we have:

- [x] separate jobs for each plugin
- [x] building against different yosys versions (`hdl/conda-eda`, HEAD, latest stable)


.github/action/action.yml
```diff
@@ -0,0 +1,76 @@
...
+    required: true
+  yosys-version:
+    description: "Which yosys version to build against ('conda' or branch name)"
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+
+    - uses: actions/checkout@v2
```
umarcor, line 31: I think this is redundant in a local Composite Action. The Composite Action requires you to checkout the repository before, because you need to provide a local path to a file that must exist. Therefore, I recommend to add the `submodules: recursive` in the `tests.yml` workflow, and remove this step here.

Alternatively, you can use a Reusable Workflow, instead of a Composite Action. The syntax is quite similar, but you will need to locate the reusable workflow in `.github/workflows/`. Then, you can call the local reusable workflow without checking out the repo. So, you would remove the step from `tests.yml` and keep it here.

Furthermore, Reusable Workflows allow you to have multiple jobs in them, not only steps.

apokusinski: I removed this checkout step from composite action and added `with: submodule: recursive' in `tests.yml`: https://github.com/chipsalliance/yosys-f4pga-plugins/pull/433/commits/64be765b636f0519c0c96b123d7a7a316c5c7837

apokusinski:
Superseded by https://github.com/chipsalliance/yosys-f4pga-plugins/pull/444 and [#445](https://github.com/chipsalliance/yosys-f4pga-plugins/pull/445)

PR #434, opened 2023-01-11
Do not duplicate ranges when bit selection operator is used

wsipak:
Related to https://github.com/antmicro/yosys-systemverilog/issues/1223

https://github.com/YosysHQ/yosys/blob/52d8ddee0ccde0afccbc703d673e46457f63672b/tests/asicworld/code_hdl_models_arbiter.v#L106-L107
This usage of the bit selection operator leads to calling `simplify` on the same wire multiple times.
In general we need to be able to repeat calls to `simplify` on wires but in this case calling it again results in creating redundant `AST_RANGE` nodes.

We've agreed that removing the `simplify` call is not the right solution, so I'm adding an attribute to acknowledge that a wire was already simplified.

I'm running tests here: https://github.com/antmicro/yosys-systemverilog/actions/runs/3891658163
With additional changes in the passlist for tests: https://github.com/antmicro/yosys-systemverilog/commits/wsip/passlist

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -67,6 +67,11 @@ enum AstNodeTypeExtended {
     static const IdString id("\\is_imported");
     return id;
 }
+/*static*/ const IdString &UhdmAst::is_simplified_wire()
+{
+    static const IdString id("\\is_simplified_wire");
+    return id;
+}
 
 static void sanitize_symbol_name(std::string &name)
 {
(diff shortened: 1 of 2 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -1633,6 +1644,7 @@ void UhdmAst::process_design()
         visitEachDescendant(current_node, [&](AST::AstNode *node) {
             node->attributes.erase(UhdmAst::packed_ranges());
             node->attributes.erase(UhdmAst::unpacked_ranges());
+            node->attributes.erase(UhdmAst::is_simplified_wire());
```
mglb, line 1647: Memory leak. Attribute nodes have to be `delete`d. `erase` just removes pointer from the map.

mglb (approved):
Just one problem, it can be merged after fixing it.
Is there a test for this issue in https://github.com/chipsalliance/UHDM-integration-tests ? If not it would be nice to add one. In case of the formal verification test failure in the future, we would be able to quickly see whether the issue came back.

mandrys (approved):

PR #444, opened 2023-01-23
Split Github Actions to be per plugin

apokusinski:
Right now there is a single Github action, in which we build and test all the plugins sequentially. This PR changes the workflow, so that there is a separate action for each plugin.


tmichalak:
LGTM

PR #445, opened 2023-01-24
CI: check different Yosys versions

apokusinski:
We would like to test the plugins builds and tests with different Yosys versions:
- Yosys in `hdl/conda-eda`
- upstream Yosys HEAD
- latest Yosys stable release

### DIFF
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -1,0 +1,71 @@
+# Copyright 2020-2022 F4PGA Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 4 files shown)


.github/actions/action.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install git g++-9 build-essential bison flex \
+          libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot \
+          pkg-config libboost-system-dev libboost-python-dev \
+          libboost-filesystem-dev zlib1g-dev clang-format-8 cmake \
+          gperf autoconf tclsh swig uuid-dev libgoogle-perftools-dev \
+          default-jre lcov
+        pip3 install orderedmultidict psutil
+    - name: Format
```
umarcor, line 44: Nitpick. Misssing newline between 43 and 44.

.github/actions/action.yml
```diff
@@ -0,0 +1,71 @@
...
+
+runs:
+  using: 'composite'
+  steps:
+
+    - uses: actions/setup-python@v2
+
+    - name: Install
+      shell: bash
+      run: |
```
umarcor, line 35: I suggest building a container in one step and reusing it for all the jobs, instead of installing all the deps in each jobs. Nevertheless, that can be done in a follow-up PR.

.github/actions/action.yml
apokusinski: Fixed in https://github.com/chipsalliance/yosys-f4pga-plugins/pull/445/commits/d5676a0ea7de1632bfa44e91c8a3bf320fc3d317

PR #447, opened 2023-01-25
Add 'final' to unsupported statements 

robertszczepanski:
`final` is unsupported by Yosys and it causes a crash with an error:
```
ERROR: :0: Encountered unhandled object '' of type 'final_stmt' 
```
From now on, there will be a warning printed since this is not a statement that should break synthesis and it is used for simulation purposes only.

FYI @mglb @wsipak @mandrys

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -4259,10 +4259,11 @@ void UhdmAst::process_primterm()
     visit_one_to_one({vpiExpr}, obj_h, [&](AST::AstNode *node) { current_node->children.push_back(node); });
 }
 
-void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object)
+void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_error)
 {
-    log_error("%.*s:%d: Currently not supported object of type '%s'\n", (int)object->VpiFile().length(), object->VpiFile().data(),
-              object->VpiLineNo(), UHDM::VpiTypeName(obj_h).c_str());
+    const auto log_func = is_error ? log_error : log_warning;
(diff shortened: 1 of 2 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -4384,6 +4384,9 @@ AST::AstNode *UhdmAst::process_object(vpiHandle obj_handle)
     case vpiInitial:
         process_initial();
         break;
+    case vpiFinal:
+        log_warning("Skipping unsupported object of type '%s'.\n", UHDM::VpiTypeName(obj_h).c_str());
```
mglb, line 4388: There is a function [`process_unsupported_stmt`](https://github.com/chipsalliance/yosys-f4pga-plugins/blob/b6a3874a3c8ff80fc3d999683db86fb7d8c1fc02/systemverilog-plugin/UhdmAst.cc#L4262-L4266)
Please modify it to add an optional argument `is_error = true` and something like:
```
const auto log_func = is_error ? log_error : log_warning;
log_func(/* arguments currently passed to log_error */);
```

robertszczepanski: I applied this change but with a little tweak. Since `final_stmt` is a `vpiProcess`, it doesn't contain info about a filename and a line number. This results in a log like:
```
Warning: :0: Currently not supported object of type 'final_stmt'
```
I made a prefix dependent on `object->VpiLineNo()` since AFAIK it will always be non-zero if processed object contains data about filename and line number.
With this change it returns:
```
Warning: Currently not supported object of type 'final_stmt'
```
And in case there's a sufficient data:
```
ERROR: /hdd/yosys-systemverilog/Surelog/tests/ContAssign/dut.sv:5: Currently not supported object of type 'cont_assign'
```
(It's of course not a real error, we do support `cont_assign`, I forced it.)

mglb: Awesome, thanks!

PR #448, opened 2023-01-31
sdc-plugin: Add missing includes.

mglb:
This file does compile (by accident) in most environments, but it did stop to compile in mine. I'm using latest yosys, so maybe that's the reason.
Anyway, the file uses things from the added includes, so the includes should be there.


tmichalak:
LGTM

PR #452, opened 2023-02-08
systemverilog-plugin: move handle of list op to process_case_item

kamilrakoczy:
Hot-fix for: https://github.com/chipsalliance/yosys-f4pga-plugins/pull/450

Ref: https://github.com/antmicro/yosys-systemverilog/issues/1448

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4125754497/jobs/7126745238

It moves list op in set membership operator from ``process_list_op`` to ``process_case_item``.

Because ``process_list_op`` adds nodes directly to ancestor previous approach breaks processes with multiple conditions.
Ultimately I think we should refactor ``process_list_op`` to return nodes to caller instead of adding it directly as it causes a lot of confusion.

I've also added assert to ``vpiOperand`` visit in ``process_case_item`` to make sure we don't add any nodes in this visitor.

Fixing this showed that opentitan also uses identifiers as list op arguments, so I've added support for this (it was in TODO in previous PR).

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3190,35 +3190,12 @@ void UhdmAst::process_list_op()
 {
     // Add all operands as children of process node
     if (auto parent_node = find_ancestor({AST::AST_ALWAYS, AST::AST_COND})) {
-        std::vector<AST::AstNode *> nodes;
-        // vpiListOp is returned in 2 cases:
-        // a, b, c ... -> multiple vpiListOp with single item
-        // [a : b] -> single vpiListOp with 2 items
         visit_one_to_many({vpiOperand}, obj_h, [&](AST::AstNode *node) {
+            // add directly to process/cond node
             if (node) {
-                nodes.push_back(node);
+                parent_node->children.push_back(node);
             }
         });
-        if (nodes.size() == 1) {
-            parent_node->children.push_back(nodes[0]);
-        } else {
-            log_assert(nodes.size() == 2);
-            // TODO(krak): we should actually simplify this nodes first,
-            // but that would require to delay this to later.
-            // For now check that they are constants.
-            log_assert(nodes[0]->type == AST::AST_CONSTANT);
-            log_assert(nodes[1]->type == AST::AST_CONSTANT);
-            const int low = nodes[0]->integer;
-            const int high = nodes[1]->integer;
-            // According to standard:
-            // If the bound to the left of the colon is greater than the
-            // bound to the right, the range is empty and contains no values.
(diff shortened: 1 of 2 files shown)


mglb (changes requested):
Please do not use `AST::Extended` assigned to `type` in new code. This causes undefined/unexpected behaviors and makes debugging harder.
Use an attribute (new one if needed) to mark node as something special.

kamilrakoczy:
Sure, I've changed to use new attribute instead.

New yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4135116919

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3510,13 +3532,27 @@ void UhdmAst::process_case_item()
...
+                log_assert(node == nullptr);
             });
+            // vpiListOp is returned in 2 cases:
+            // a, b, c ... -> multiple vpiListOp with single item
+            // [a : b] -> single vpiListOp with 2 items
+            // single item is handled by default,
+            // here handle 2 items with custom low_high_bound attribute
+            if (current_node->children.size() == 2) {
+                auto block = make_ast_node(AST::AST_BLOCK);
+                block->children = current_node->children;
```
mglb, line 3551:
```suggestion
                block->children = std::move(current_node->children);
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1028,6 +1034,45 @@ static void simplify(AST::AstNode *current_node, AST::AstNode *parent_node)
...
+            // If the bound to the left of the colon is greater than the
+            // bound to the right, the range is empty and contains no values.
+            for (int i = low; i >= low && i <= high; i++) {
+                // TODO(krak): what should be correct constant range if we don't have
+                // information about range?
+                if (range == -1) {
+                    range = 64;
+                }
+                current_node->children.push_back(AST::AstNode::mkconst_int(i, false, range));
+            }
```
mglb, line 1071: I believe `range_left == -1` is not an absolute indicator of invalid range. Shouldn't we check `range_valid`? Also, this ignores the range of `high`.

Also, the default should be 32:
- 32 seems to be Yosys default int constant size,
- you're storing the constant in an `int` here, so anything bigger won't work anyway,
- `mkconst_int` takes 32 bit value, so again, anything bigger won't work.

mglb (changes requested):

kamilrakoczy:
New yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4162797345/jobs/7202408725

mglb (approved):

PR #454, opened 2023-02-13
Handle type parameters

wsipak:
Depends on https://github.com/chipsalliance/yosys-f4pga-plugins/pull/463 and https://github.com/chipsalliance/yosys-f4pga-plugins/pull/468
When a module with `type parameter` is instantiated, Surelog replaces the paramater with proper type name.
However, if the type is not globally defined, the `typedef` for that type is missing in the parametrized module.
The plugin should copy the typedef into proper scope.

New tests:
https://github.com/chipsalliance/UHDM-integration-tests/pull/707

CI with the changes:
https://github.com/antmicro/yosys-systemverilog/actions/runs/4618129458

CI with the changes and the new tests:
https://github.com/antmicro/yosys-systemverilog/actions/runs/4618128622

CI with the new tests and `main` plugins:
https://github.com/antmicro/yosys-systemverilog/actions/runs/4608097188

Test `yosys:opt/opt_lut.v` starts to fail when the new IdString is added, I'm ignoring it here.
Similarly with `yosys:simple/multiplier.v`.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -53,6 +53,7 @@ static IdString force_convert;
 static IdString is_imported;
 static IdString is_simplified_wire;
 static IdString low_high_bound;
+static IdString is_type_parameter;
 }; // namespace attr_id
 
 // TODO(mglb): use attr_id::* directly everywhere and remove those methods.
@@ -83,6 +84,7 @@ void attr_id_init()
     attr_id::is_imported = IdString("$systemverilog_plugin$is_imported");
     attr_id::is_simplified_wire = IdString("$systemverilog_plugin$is_simplified_wire");
     attr_id::low_high_bound = IdString("$systemverilog_plugin$low_high_bound");
+    attr_id::is_type_parameter = IdString("$systemverilog_plugin$is_type_parameter");
 }
 
 void attr_id_cleanup()
@@ -95,6 +97,25 @@ void attr_id_cleanup()
     attr_id::unpacked_ranges = IdString();
     attr_id::packed_ranges = IdString();
     attr_id::partial = IdString();
+    attr_id::is_type_parameter = IdString();
+}
+
+static AST::AstNode *get_attribute(AST::AstNode *node, const IdString &attribute)
+{
+    log_assert(node);
+    if (!node->attributes.count(attribute))
+        return nullptr;
+
+    return node->attributes[attribute];
+}
+
+// Consumes attr_node.
+static void set_attribute(AST::AstNode *node, const IdString &attribute, AST::AstNode *attr_node)
+{
+    log_assert(node);
+    log_assert(attr_node);
+    delete node->attributes[attribute];
+    node->attributes[attribute] = attr_node;
 }
 
 // Delete the selected attribute if it exists.
@@ -118,7 +139,7 @@ static void delete_internal_attributes(AST::AstNode *node)
         return;
 
     for (auto &attr : {UhdmAst::partial(), UhdmAst::packed_ranges(), UhdmAst::unpacked_ranges(), UhdmAst::force_convert(), UhdmAst::is_imported(),
-                       UhdmAst::is_simplified_wire(), UhdmAst::low_high_bound()}) {
+                       UhdmAst::is_simplified_wire(), UhdmAst::low_high_bound(), attr_id::is_type_parameter}) {
         delete_attribute(node, attr);
     }
 }
@@ -1943,6 +1964,7 @@ void UhdmAst::process_module()
 
             delete_attribute(current_node, UhdmAst::partial());
         } else {
+            // processing nodes belonging to 'uhdmallModules'
             current_node = make_ast_node(AST::AST_MODULE);
             current_node->str = type;
             shared.top_nodes[current_node->str] = current_node;
@@ -1955,7 +1977,8 @@ void UhdmAst::process_module()
             });
             visit_one_to_many({vpiModule, vpiParameter, vpiParamAssign, vpiNet, vpiArrayNet, vpiProcess}, obj_h, [&](AST::AstNode *node) {
                 if (node) {
-                    if (node->type == AST::AST_ASSIGN && node->children.size() < 2) {
+                    if ((node->type == AST::AST_ASSIGN && node->children.size() < 2) ||
+                        (node->type == AST::AST_PARAMETER && get_attribute(node, attr_id::is_type_parameter))) {
                         delete node;
                         return;
                     }
@@ -1967,6 +1990,42 @@ void UhdmAst::process_module()
         // Not a top module, create instance
         current_node = make_ast_node(AST::AST_CELL);
         std::vector<std::pair<RTLIL::IdString, RTLIL::Const>> parameters;
+
+        std::vector<AST::AstNode *> parameter_typedefs;
+
+        visit_one_to_many({vpiParameter}, obj_h, [&](AST::AstNode *node) {
+            log_assert(node);
+            AST::AstNode *attr = get_attribute(node, attr_id::is_type_parameter);
+
+            if (!attr) {
+                // Process type parameters only.
+                delete node;
+                return;
+            }
+
+            if (node->children.size() == 0) {
+                log_assert(!attr->str.empty());
+                // Anonymous types have no chidren, and store the parameter name in attr->str.
+                parameters.push_back(std::make_pair(node->str, attr->str));
+                delete node;
+                return;
+            }
+
+            for (auto child : node->children) {
+                if (child->type == AST::AST_TYPEDEF && !child->str.empty()) {
+                    // process_type_parameter should have created a node with the parameter name
+                    //   and a child with the name of the value assigned to the parameter.
+                    parameters.push_back(std::make_pair(node->str, child->str));
+                }
+
+                if (child->type == AST::AST_TYPEDEF || child->type == AST::AST_ENUM) {
+                    // Copy definition of the type provided as parameter.
+                    parameter_typedefs.push_back(child->clone());
+                }
+            }
+            delete node;
+        });
+
         visit_one_to_many({vpiParamAssign}, obj_h, [&](AST::AstNode *node) {
             if (node && node->type == AST::AST_PARAMETER) {
                 log_assert(!node->children.empty());
@@ -2023,6 +2082,7 @@ void UhdmAst::process_module()
                 }
             }
         });
+        module_node->children.insert(std::end(module_node->children), std::begin(parameter_typedefs), std::end(parameter_typedefs));
         if (module_node->attributes.count(UhdmAst::partial())) {
             AST::AstNode *attr = module_node->attributes.at(UhdmAst::partial());
             if (attr->type == AST::AST_CONSTANT)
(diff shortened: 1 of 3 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -4540,6 +4594,83 @@ void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_er
...
+        }
+    }
+    if (current_node->children.size() == 0) {
+        // nothing to do with the parameter, the actual value was substituted everywhere already in uhdm
+        if (current_node->attributes.count(attr_id::is_type_parameter)) {
+            current_node->attributes.erase(attr_id::is_type_parameter);
+        }
+        delete current_node;
+        current_node = nullptr;
+    }
```
kamilrakoczy, line 4671: I think we should error out when we can't find parameter type, otherwise we could wrongly rename module.
```suggestion
    if (current_node->children.size() == 0) {
        log_file_error(current_node->filename, current_node->location.first_line, "Empty parameter type!");
    }
```

kamilrakoczy (changes requested):
My biggest concern is about empty type parameters. I think we shouldn't allow them. If this means that we need to support anonymous type parameters, I think we should do this.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -83,6 +84,7 @@ void attr_id_init()
     attr_id::is_imported = IdString("$systemverilog_plugin$is_imported");
     attr_id::is_simplified_wire = IdString("$systemverilog_plugin$is_simplified_wire");
     attr_id::low_high_bound = IdString("$systemverilog_plugin$low_high_bound");
+    attr_id::is_type_parameter = IdString("$systemverilog_plugin$is_type_parameter");
```
kamilrakoczy, line 87: Please add this parameter to ``delete_internal_attributes``

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1853,6 +1874,7 @@ void UhdmAst::process_design()
             continue;
         if (pair.second->type == AST::AST_PACKAGE) {
             check_memories(pair.second);
+            clear_current_scope();
```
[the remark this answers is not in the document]
wsipak: Alright, for now I've moved it to another commit.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -95,6 +97,25 @@ void attr_id_cleanup()
...
+    log_assert(node);
+    if (!node->attributes.count(attribute))
+        return nullptr;
+
+    return node->attributes[attribute];
+}
+
+static void set_attribute(AST::AstNode *node, const IdString &attribute, AST::AstNode *attr_node)
+{
+    // This function consumes attr_node.
```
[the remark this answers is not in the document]

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1919,6 +1941,7 @@ void UhdmAst::process_module()
     name = strip_package_name(name);
     if (!is_module_instance) {
         if (shared.top_nodes.find(type) != shared.top_nodes.end()) {
+            // processing nodes belonging to 'uhdmtopModules'
```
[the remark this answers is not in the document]
wsipak: I didn't realise it wasn't accurate. I've removed the comment as it might be confusing in case it's not only `uhdmtopModules`.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1966,6 +1990,35 @@ void UhdmAst::process_module()
         // Not a top module, create instance
         current_node = make_ast_node(AST::AST_CELL);
         std::vector<std::pair<RTLIL::IdString, RTLIL::Const>> parameters;
+
+        std::vector<AST::AstNode *> parameter_typedefs;
+
+        // For some reason 'vpiParameter' causes 'vpiTypeParameter' to be found, too.
+        visit_one_to_many({vpiParameter}, obj_h, [&](AST::AstNode *node) {
+            if (!node)
+                return;
```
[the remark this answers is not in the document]
wsipak: Alright, so we're going to support anonymous types and the `node` is asserted now.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1966,6 +1990,35 @@ void UhdmAst::process_module()
...
+                // process type parameters only
+                delete node;
+                return;
+            }
+
+            for (auto child : node->children) {
+                if (child->type == AST::AST_TYPEDEF && !child->str.empty()) {
+                    // process_type_parameter should have created a node with the parameter name
+                    //   and a child with the name of the value assigned to the parameter
+                    parameters.push_back(std::make_pair(node->str, child->str));
```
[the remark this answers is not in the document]
wsipak: No longer relevant, as we have the changes to support anonymous types as parameters.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -4540,6 +4594,83 @@ void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_er
...
+    set_attribute(current_node, attr_id::is_type_parameter, AST::AstNode::mkconst_int(1, false, 1));
+    std::string renamed_enum = "";
+
+    visit_one_to_one({vpiTypespec}, obj_h, [&](AST::AstNode *node) {
+        if (!node)
+            return;
+
+        if (node->type == AST::AST_WIRE && node->str == "") {
+            // skip AST_WIREs that don't have a name,
+            // this happens when processing default values of parameters
```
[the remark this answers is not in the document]
wsipak: Done, anonymous types affect the name of the module now.

systemverilog-plugin/UhdmAst.cc
wsipak: I've introduced changes to support anonymous types, and this is no longer relevant.
Parameters with anonymous types do have 0 children but now this is allowed, as the `is_type_parameter` attribute is used and that's enough.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -4547,6 +4607,73 @@ void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_er
     log_func("%sCurrently not supported object of type '%s'\n", prefix.c_str(), UHDM::VpiTypeName(obj_h).c_str());
 }
 
+void UhdmAst::process_type_parameter()
+{
+    current_node = make_ast_node(AST::AST_PARAMETER);
+    static int anonymous_type_id = 0;
+    static int enum_id = 0;
```
kamilrakoczy, line 4614: Please add this static variables to ``uhdmastshared``. It also looks like there is already some static for enum: https://github.com/chipsalliance/yosys-f4pga-plugins/blob/main/systemverilog-plugin/uhdmastshared.h#L32

Can we use it?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -4547,6 +4607,73 @@ void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_er
...
+
+    // Use an attribute to distinguish "type parameters" from other parameters
+    set_attribute(current_node, attr_id::is_type_parameter, AST::AstNode::mkconst_int(1, false, 1));
+    std::string renamed_enum = "";
```
kamilrakoczy, line 4618:
```suggestion
    std::string renamed_enum;
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -4547,6 +4607,73 @@ void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_er
...
+                if (child->str == node->str && child->type == AST::AST_TYPEDEF) {
+                    current_node->children.push_back(child->clone());
+                    break;
+                }
+            }
+        }
+        delete node;
+    });
+
+    if (renamed_enum != "") {
```
kamilrakoczy, line 4659:
```suggestion
    if (!renamed_enum.empty()) {
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -4547,6 +4607,73 @@ void UhdmAst::process_unsupported_stmt(const UHDM::BaseClass *object, bool is_er
...
+
+    // Use an attribute to distinguish "type parameters" from other parameters
+    set_attribute(current_node, attr_id::is_type_parameter, AST::AstNode::mkconst_int(1, false, 1));
+    std::string renamed_enum = "";
+
+    visit_one_to_one({vpiTypespec}, obj_h, [&](AST::AstNode *node) {
+        if (!node)
+            return;
+
+        if (node->type == AST::AST_WIRE && node->str == "") {
```
kamilrakoczy, line 4624:
```suggestion
        if (node->type == AST::AST_WIRE && node->str.empty()) {
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -4824,6 +4951,7 @@ AST::AstNode *UhdmAst::process_object(vpiHandle obj_handle)
         // so the plugin doesn't need to process the parameter itself.
         // Other parameter types are handled by the plugin
         // mainly because they were implemented before Surelog did the substitution.
+        process_type_parameter();
```
kamilrakoczy, line 4954: Is above comment still relevant? I think now we are processing type parameters, so it is no longer valid.

kamilrakoczy (approved):
LGTM, only minor suggestions.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1966,6 +1990,41 @@ void UhdmAst::process_module()
...
+            for (auto child : node->children) {
+                if (child->type == AST::AST_TYPEDEF && !child->str.empty()) {
+                    // process_type_parameter should have created a node with the parameter name
+                    //   and a child with the name of the value assigned to the parameter.
+                    parameters.push_back(std::make_pair(node->str, child->str));
+                }
+
+                if (child->type == AST::AST_TYPEDEF || child->type == AST::AST_ENUM) {
+                    // Copy definition of the type provided as parameter.
+                    parameter_typedefs.push_back(child->clone());
```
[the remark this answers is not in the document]
wsipak: `module_node` is retrieved later on in the code, so we can't use it here, and parameter handling has to be done before retrieving `module_node`, so I needed to create an additional vector for storing these objects.

systemverilog-plugin/UhdmAst.cc
wsipak: That's right, now we're using the existing enum counter, and the counter for anonymous types works analogously to what we've already had in `uhdmastshared`.

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc
wsipak: I agree this is no longer valid. I've removed the comment.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1966,6 +1990,41 @@ void UhdmAst::process_module()
...
+            log_assert(node);
+            AST::AstNode *attr = get_attribute(node, attr_id::is_type_parameter);
+
+            if (!attr) {
+                // Process type parameters only.
+                delete node;
+                return;
+            }
+
+            if (node->children.size() == 0 && attr->str != "") {
```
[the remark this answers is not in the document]
wsipak: Now I see this could be confusing. I've added a `log_assert` and a comment explaining that we expect `attr->str` not to be empty here.

systemverilog-plugin/UhdmAst.cc

mandrys (approved):

PR #455, opened 2023-02-15
systemverilog-plugin: prefer full name in function calls

kamilrakoczy:
This PR fixes usage of imported function that uses another, not imported, function.

UHDM-integration-tests: https://github.com/chipsalliance/UHDM-integration-tests/pull/708
yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4183705520

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1676,13 +1676,12 @@ AST::AstNode *UhdmAst::find_ancestor(const std::unordered_set<AST::AstNodeType> 
 void UhdmAst::process_design()
 {
     current_node = make_ast_node(AST::AST_DESIGN);
-    visit_one_to_many(
-      {UHDM::uhdmallInterfaces, UHDM::uhdmallPackages, UHDM::uhdmtopPackages, UHDM::uhdmallModules, UHDM::uhdmtopModules, vpiTaskFunc}, obj_h,
-      [&](AST::AstNode *node) {
-          if (node) {
-              shared.top_nodes[node->str] = node;
-          }
-      });
+    visit_one_to_many({UHDM::uhdmallInterfaces, UHDM::uhdmtopPackages, UHDM::uhdmallModules, UHDM::uhdmtopModules, vpiTaskFunc}, obj_h,
+                      [&](AST::AstNode *node) {
+                          if (node) {
+                              shared.top_nodes[node->str] = node;
+                          }
+                      });
     visit_one_to_many({vpiParameter, vpiParamAssign}, obj_h, [&](AST::AstNode *node) {});
     visit_one_to_many({vpiTypedef}, obj_h, [&](AST::AstNode *node) {
         if (node)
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -3824,6 +3824,12 @@ void UhdmAst::process_tf_call(AST::AstNodeType type)
             current_node->children.push_back(node);
         }
     });
+    // prefer full name of function
+    // this fixes case when we imports single function from package
+    // but it calls another, not imported, function
+    if (vpiHandle function_h = vpi_handle(vpiFunction, obj_h)) {
+        current_node->str = get_name(function_h, true);
```
wsipak, line 3831: Is it possible that we could have the same result by using `make_ast_node` with the argument `prefer_full_name` set to `true`?
```
AST::AstNode *UhdmAst::make_ast_node(AST::AstNodeType type, std::vector<AST::AstNode *> children, bool prefer_full_name)
```
It seems to that this way we would have the proper name set immediately, without the need to modify `node->str` after the node object is created.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3824,6 +3824,12 @@ void UhdmAst::process_tf_call(AST::AstNodeType type)
             current_node->children.push_back(node);
         }
     });
+    // prefer full name of function
+    // this fixes case when we imports single function from package
```
wsipak, line 3828: I think it should be `when we import a single function from a package`. Or I could suggest an impersonal alternative: `When a single function is imported from a package`

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: Unfortunately not, ``vpiFuncCall`` have only name without scope, but ``vpiFunction`` have full name:
```
       \_return_stmt: , line:7:5, endln:7:11
         |vpiParent:
         \_function: (pkg::func_1), line:6:3, endln:8:14
         |vpiCondition:
         \_func_call: (func_2), line:7:12, endln:7:20
           |vpiParent:
           \_return_stmt: , line:7:5, endln:7:11
           |vpiName:func_2
           |vpiFunction:
           \_function: (pkg::func_2), line:3:3, endln:5:14
 ```

File: systemverilog-plugin/UhdmAst.cc (hunk 2 above)
Comment (kamilrakoczy) [reply systemverilog-plugin/UhdmAst.cc]: Sure, changed.

**Review by wsipak — APPROVED** [review]

File: systemverilog-plugin/UhdmAst.cc [hunk 3]
```diff
@@ -3825,6 +3824,12 @@ void UhdmAst::process_tf_call(AST::AstNodeType type)
             current_node->children.push_back(node);
         }
     });
+    // prefer full name of function, with scope
+    // important, when a single function is imported from a package
+    // but it calls another, not imported, function
```
mglb, line 3829:
```suggestion
    // Prefer fully qualified name of a function (prefixed with a scope).
    // This is important when a single function which has been imported from a package
    // calls another function that is not imported in the calling scope.
```
Did I understand the message correctly?

mglb (approved):
The code looks OK. I've proposed a change in the comment to make it a bit more clear.

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: Yes, I've updated comment, it looks better.

PR #458, opened 2023-02-17
systemverilog-plugin: convert memory when accessing slice

kamilrakoczy:
Fixes: https://github.com/antmicro/yosys-systemverilog/issues/1505

Convert memories to list of registers if we are accessing slice of memory.

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4204594665
UHDM-integration test: https://github.com/chipsalliance/UHDM-integration-tests/pull/711

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -439,9 +439,15 @@ static void check_memories(AST::AstNode *module_node)
                 //   mem <= '{default:0};
                 //   mem[7:1] <= mem[6:0];
                 // end
+                // don't convert in case of accessing
+                // memory using address, e.g.
+                // mem[0] <= '{default:0}
+                //
+                // Access to whole memory
                 if (node->children.size() == 0) {
                     force_convert = true;
                 }
+                // Access to slice of memory
                 if (node->children.size() == 1 && node->children[0]->children.size() != 1) {
                     force_convert = true;
                 }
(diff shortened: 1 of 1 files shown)


mandrys (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -428,8 +428,26 @@ static void check_memories(AST::AstNode *module_node)
...
+                // always_ff @ (posedge clk) begin
+                //   mem <= '{default:0};
+                //   mem[7:1] <= mem[6:0];
+                // end
+                if (node->children.size() == 0) {
+                    force_convert = true;
+                }
+                if (node->children.size() == 1 && node->children[0]->children.size() != 1) {
+                    force_convert = true;
+                }
```
wsipak, line 447: It's unclear to me what these `if`s really do. From what I see, it could either be an identifier with no children, or identifier with one range having both a beginning and an end?
We could have more comments explaining the decision.

kamilrakoczy: Yes, you are right, it is identifier with no children (if someone access whole memory), or identifier with one range having both a beginning and an end (if someone access slice of memory).

I've added this to comment and added information what type of memories we don't want to convert.

wsipak: OK, thank you.

wsipak (approved):

PR #460, opened 2023-02-22
systemverilog-plugin: handle assignment pattern with only default inside cell port

kamilrakoczy:
This PR adds support for assignment patters inside cell high conn  port that is setting all values to come constant, e.g.:

```
mod  m (.data( '{default: '0} ));
```

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3716,8 +3716,22 @@ void UhdmAst::process_tagged_pattern()
         lhs_node = assign_node->children[0];
     } else {
         lhs_node = new AST::AstNode(AST::AST_IDENTIFIER);
-        auto ancestor = find_ancestor({AST::AST_WIRE, AST::AST_MEMORY, AST::AST_PARAMETER, AST::AST_LOCALPARAM});
-        if (!ancestor) {
+        auto ancestor = find_ancestor({AST::AST_WIRE, AST::AST_MEMORY, AST::AST_PARAMETER, AST::AST_LOCALPARAM, AST::AST_CELL});
+        if (ancestor == AST::AST_CELL) {
+            // special case: if this assignment pattern is inside cell high conn port
+            // and we are setting all values to some constant
+            auto typespec_h = vpi_handle(vpiTypespec, obj_h);
+            if (typespec_h && vpi_get(vpiType, typespec_h) == vpiStringTypespec) {
+                std::string field_name = vpi_get_str(vpiName, typespec_h);
+                if (field_name == "default") {
+                    visit_one_to_one({vpiPattern}, obj_h, [&](AST::AstNode *node) { current_node = node; });
+                    vpi_release_handle(typespec_h);
+                    return;
+                }
+            }
+            vpi_release_handle(typespec_h);
+            log_error("Couldn't find ancestor for tagged pattern!\n");
+        } else if (!ancestor) {
             log_error("Couldn't find ancestor for tagged pattern!\n");
         }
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -3713,17 +3713,31 @@ void UhdmAst::process_tagged_pattern()
...
     if (assign_node) {
         assign_type = assign_node->type;
-        lhs_node = assign_node->children[0];
+        lhs_node = assign_node->children[0]->clone();
     } else {
         lhs_node = new AST::AstNode(AST::AST_IDENTIFIER);
         auto ancestor = find_ancestor({AST::AST_WIRE, AST::AST_MEMORY, AST::AST_PARAMETER, AST::AST_LOCALPARAM});
         if (!ancestor) {
+            // special case: if this assignment pattern is inside cell high conn port
+            if (find_ancestor({AST::AST_CELL})) {
```
mglb, line 3722: Is it possible to add `AST::AST_CELL` to `find_ancestor` 3 lines above and reorganize the condition into something like:
```
if (!ancestor) {
    log_error(...);
}
if (ancestor->type == AST::AST_CELL) {
    ...
}
```
Reason: less nesting and one `find_ancestor` less.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3713,17 +3713,31 @@ void UhdmAst::process_tagged_pattern()
     AST::AstNode *lhs_node = nullptr;
     if (assign_node) {
         assign_type = assign_node->type;
-        lhs_node = assign_node->children[0];
+        lhs_node = assign_node->children[0]->clone();
     } else {
         lhs_node = new AST::AstNode(AST::AST_IDENTIFIER);
         auto ancestor = find_ancestor({AST::AST_WIRE, AST::AST_MEMORY, AST::AST_PARAMETER, AST::AST_LOCALPARAM});
```
mglb, line 3719: While you're there: `const auto* const`.
Same above, in `auto assign_node = ...`. This change will actually make `->clone()` you've added mandatory.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3713,17 +3713,31 @@ void UhdmAst::process_tagged_pattern()
...
         if (!ancestor) {
+            // special case: if this assignment pattern is inside cell high conn port
+            if (find_ancestor({AST::AST_CELL})) {
+                auto typespec_h = vpi_handle(vpiTypespec, obj_h);
+                if (typespec_h && vpi_get(vpiType, typespec_h) == vpiStringTypespec) {
+                    // and we are setting all values to some constant
```
mglb, line 3725:
```suggestion
                // and we are setting all values to some constant
                if (typespec_h && vpi_get(vpiType, typespec_h) == vpiStringTypespec) {
```
For consistency with the comment + if above

mglb (changes requested):
Please also add some minimal test to uhdm-integration-tests.
Such test will allow detecting regressions in this code directly. Otherwise we would need to dig through OpenTitan logs to find out what exactly failed.

wsipak (approved):

kamilrakoczy:
UHDM-integration-tests: https://github.com/chipsalliance/UHDM-integration-tests/pull/714

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4281547440

kamilrakoczy:
After: https://github.com/chipsalliance/Surelog/pull/3486 Surelog started to flattening such assignments, so this PR is no longer needed. Closing.

PR #461, opened 2023-02-22
Memory in nested genblocks

mandrys:
These changes add support for memories in nested genblocks, where we need to track the current scope. Before this fix, memories with same names but in different scopes (inside and outside a genblock) were not distinguished in `check_memories` that is called to add `force_convert` attribute when we need to convert a memory into an array.
This issue was spotted when parsing OpenTitan design, after getting the following error:
```
ERROR: Assert `!memories.count(node->str)' failed in /yosys-systemverilog/yosys-f4pga-plugins/systemverilog-plugin/UhdmAst.cc:419.
```
Here is a test for the fix: https://github.com/chipsalliance/UHDM-integration-tests/pull/712,
and the PR with test workflow: https://github.com/antmicro/yosys-systemverilog/pull/1528.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -399,42 +399,61 @@ static void resolve_wiretype(AST::AstNode *wire_node)
     }
 }
 
-static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
+static void add_force_convert_attribute(AST::AstNode *wire_node, uint32_t val = 1)
 {
-    wire_node->attributes[UhdmAst::force_convert()] = AST::AstNode::mkconst_int(val, true);
+    AST::AstNode *&attr = wire_node->attributes[UhdmAst::force_convert()];
+    if (!attr) {
+        attr = AST::AstNode::mkconst_int(val, true);
+    } else if (attr->integer != val) {
+        attr->integer = val;
+    }
 }
 
-static void check_memories(AST::AstNode *module_node)
+static void check_memories(AST::AstNode *node, std::string scope, std::map<std::string, AST::AstNode *> &memories)
 {
-    std::map<std::string, AST::AstNode *> memories;
-    std::string current_scope_str = "";
-    visitEachDescendant(module_node, [&](AST::AstNode *node) {
-        if (node->type == AST::AST_GENBLOCK) {
-            current_scope_str = "." + node->str;
+    for (auto *child : node->children) {
+        check_memories(child, node->type == AST::AST_GENBLOCK ? scope + "." + node->str : scope, memories);
+    }
+
+    if (node->str == "\\$readmemh") {
+        if (node->children.size() != 2 || node->children[1]->str.empty() || node->children[1]->type != AST::AST_IDENTIFIER) {
+            log_error("%s:%d: Wrong usage of '\\$readmemh'\n", node->filename.c_str(), node->location.first_line);
         }
-        if (node->str == "\\$readmemh") {
-            if (node->children.size() != 2 || node->children[1]->str.empty() || node->children[1]->type != AST::AST_IDENTIFIER) {
-                log_error("%s:%d: Wrong usage of '\\$readmemh'\n", node->filename.c_str(), node->location.first_line);
-            }
-            if (memories[current_scope_str + node->children[1]->str])
-                add_force_convert_attribute(memories[node->children[1]->str], 0);
+        // TODO: Look for the memory in all other scope levels, like we do in case of AST::AST_IDENTIFIER,
+        // as here the memory can also be defined before before the current scope.
+        std::string name = scope + "." + node->children[1]->str;
+        const auto iter = memories.find(name);
+        if (iter != memories.end()) {
+            add_force_convert_attribute(iter->second, 0);
         }
-        if (node->type == AST::AST_WIRE) {
-            const std::vector<AST::AstNode *> packed_ranges =
-              node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children : std::vector<AST::AstNode *>();
-            const std::vector<AST::AstNode *> unpacked_ranges = node->attributes.count(UhdmAst::unpacked_ranges())
-                                                                  ? node->attributes[UhdmAst::unpacked_ranges()]->children
-                                                                  : std::vector<AST::AstNode *>();
-            if (packed_ranges.size() == 1 && unpacked_ranges.size() == 1) {
-                log_assert(!memories.count(current_scope_str + node->str));
-                memories[current_scope_str + node->str] = node;
-            }
+    }
+
+    if (node->type == AST::AST_WIRE) {
+        const std::size_t packed_ranges_count =
+          node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children.size() : 0;
+        const std::size_t unpacked_ranges_count =
+          node->attributes.count(UhdmAst::unpacked_ranges()) ? node->attributes[UhdmAst::unpacked_ranges()]->children.size() : 0;
+
+        if (packed_ranges_count == 1 && unpacked_ranges_count == 1) {
+            std::string name = scope + "." + node->str;
+            auto [iter, did_insert] = memories.insert_or_assign(std::move(name), node);
+            log_assert(did_insert);
         }
-        if (node->type == AST::AST_IDENTIFIER) {
-            std::string name = current_scope_str + node->str;
-            bool force_convert = false;
-            if (memories.count(name)) {
-                if (!memories[name]->attributes.count(UhdmAst::force_convert())) {
+        return;
+    }
+
+    if (node->type == AST::AST_IDENTIFIER) {
+        std::string full_id = scope;
+        std::size_t scope_end_pos = scope.size();
+
+        for (;;) {
+            full_id += "." + node->str;
+            const auto iter = memories.find(full_id);
+            if (iter != memories.end()) {
+                // Memory node found!
+                if (!iter->second->attributes.count(UhdmAst::force_convert())) {
+                    const bool is_full_memory_access = (node->children.size() == 0);
+                    const bool is_slice_memory_access = (node->children.size() == 1 && node->children[0]->children.size() != 1);
                     // convert memory to list of registers
                     // in case of access to whole memory
                     // or slice of memory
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
-                // always_ff @ (posedge clk) begin
-                //   mem <= '{default:0};
-                //   mem[7:1] <= mem[6:0];
-                // end
-                // don't convert in case of accessing
-                // memory using address, e.g.
-                // mem[0] <= '{default:0}
-                //
-                // Access to whole memory
+        // 5. If we don't find any scope, we check if the memory is declared in the main scope.
```
kamilrakoczy, line 477: Maybe lets start with empty scope name, so we won't need this.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+            log_assert(!memories.count(name));
+            memories[name] = node;
+        }
+    }
+    if (node->type == AST::AST_IDENTIFIER) {
+        std::string name = "";
+        bool force_convert = false;
+        name = "." + node->str;
+        // 1. We iterate through the list of memories.
+        for (auto m : memories) {
```
kamilrakoczy, line 453: ``memories`` are ``std::map``, maybe we can just check if this id is inside this map?.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
-                memories[node->str] = node;
+    std::string full_current_scope = "";
+    std::size_t mem;
+    bool mem_found = false;
+
+    if (node->type == AST::AST_GENBLOCK) {
+        scopes.push_back(node->str);
+    }
+
+    for (auto s : scopes) {
```
kamilrakoczy, line 419: Maybe lets already store full scope in scopes, so this won't be needed?

kamilrakoczy (changes requested):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
-                    add_force_convert_attribute(memories[node->str]);
-                }
             }
         }
-    });
+        if (force_convert) {
+            add_force_convert_attribute(memories[name]);
+        }
+    }
+    return memories;
```
mglb, line 489: Just pass memories as a reference, without returning it.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+    for (auto child : node->children) {
+        memories = check_memories(child, scopes, memories);
+    }
+
+    if (node->str == "\\$readmemh") {
+        if (node->children.size() != 2 || node->children[1]->str.empty() || node->children[1]->type != AST::AST_IDENTIFIER) {
+            log_error("%s:%d: Wrong usage of '\\$readmemh'\n", node->filename.c_str(), node->location.first_line);
+        }
+        std::string mem_name = full_current_scope + "." + node->children[1]->str;
+        if (memories[mem_name]) {
```
mglb, line 433: This inserts an element when it does not exist. Use https://en.cppreference.com/w/cpp/container/map/find and use returned iterator below instead of `memories[mem_name]`

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+    }
+    if (node->type == AST::AST_WIRE) {
+        const std::vector<AST::AstNode *> packed_ranges =
+          node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children : std::vector<AST::AstNode *>();
+        const std::vector<AST::AstNode *> unpacked_ranges =
+          node->attributes.count(UhdmAst::unpacked_ranges()) ? node->attributes[UhdmAst::unpacked_ranges()]->children : std::vector<AST::AstNode *>();
+        if (packed_ranges.size() == 1 && unpacked_ranges.size() == 1) {
+            std::string name = full_current_scope + "." + node->str;
+            log_assert(!memories.count(name));
+            memories[name] = node;
```
mglb, line 445:
```suggestion
            const auto mememory_it = memories.find(name);
            log_assert(!memory_it != memories.end());
            memory_it->second = node;
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+            add_force_convert_attribute(memories[mem_name], 0);
+        }
+    }
+    if (node->type == AST::AST_WIRE) {
+        const std::vector<AST::AstNode *> packed_ranges =
+          node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children : std::vector<AST::AstNode *>();
+        const std::vector<AST::AstNode *> unpacked_ranges =
+          node->attributes.count(UhdmAst::unpacked_ranges()) ? node->attributes[UhdmAst::unpacked_ranges()]->children : std::vector<AST::AstNode *>();
+        if (packed_ranges.size() == 1 && unpacked_ranges.size() == 1) {
+            std::string name = full_current_scope + "." + node->str;
```
mglb, line 443:
```suggestion
            const std::string name = full_current_scope + "." + node->str;
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+        if (memories[mem_name]) {
+            add_force_convert_attribute(memories[mem_name], 0);
+        }
+    }
+    if (node->type == AST::AST_WIRE) {
+        const std::vector<AST::AstNode *> packed_ranges =
+          node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children : std::vector<AST::AstNode *>();
+        const std::vector<AST::AstNode *> unpacked_ranges =
+          node->attributes.count(UhdmAst::unpacked_ranges()) ? node->attributes[UhdmAst::unpacked_ranges()]->children : std::vector<AST::AstNode *>();
+        if (packed_ranges.size() == 1 && unpacked_ranges.size() == 1) {
```
mglb, line 442:
```suggestion
        const std::size_t packed_ranges_size =
          node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children.size() : 0;
        const std::size_t unpacked_ranges_size =
          node->attributes.count(UhdmAst::unpacked_ranges()) ? node->attributes[UhdmAst::unpacked_ranges()]->children.size() : 0;
        if (packed_ranges_size == 1 && unpacked_ranges_size == 1) {
```

mglb (changes requested):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+    std::string full_current_scope = "";
+    std::size_t mem;
+    bool mem_found = false;
+
+    if (node->type == AST::AST_GENBLOCK) {
+        scopes.push_back(node->str);
+    }
+
+    for (auto s : scopes) {
+        full_current_scope = full_current_scope + "." + s;
```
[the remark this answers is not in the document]
wsipak: Ah, nevermind, now I see there's no need to get rid of the dot and it's just a temporary string.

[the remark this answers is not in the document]
mandrys: Yeah, that's rigth

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+                            if (node->children.size() == 0) {
+                                force_convert = true;
+                            }
+                        }
+                        break;
+                    }
+                }
+                // 4. If we find the scope, we end serching through the list of memories.
+                if (mem_found)
+                    break;
```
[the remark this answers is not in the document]
mandrys: It is in fact needed to break another `for` in which we are nested.

systemverilog-plugin/UhdmAst.cc
mandrys: We can't have the node name as a key in the memories map, as it has to be unique for each element, and we could have multiple memories with same node name but in different scopes, so we need to have the full scope name as a key in the map.

systemverilog-plugin/UhdmAst.cc
mandrys: Right, good point.

mandrys: I reworked the changes, and now we work on a scope string only instead of a list of scopes.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -404,58 +405,88 @@ static void add_force_convert_attribute(AST::AstNode *wire_node, int val = 1)
...
+          node->attributes.count(UhdmAst::packed_ranges()) ? node->attributes[UhdmAst::packed_ranges()]->children : std::vector<AST::AstNode *>();
+        const std::vector<AST::AstNode *> unpacked_ranges =
+          node->attributes.count(UhdmAst::unpacked_ranges()) ? node->attributes[UhdmAst::unpacked_ranges()]->children : std::vector<AST::AstNode *>();
+        if (packed_ranges.size() == 1 && unpacked_ranges.size() == 1) {
+            std::string name = full_current_scope + "." + node->str;
+            log_assert(!memories.count(name));
+            memories[name] = node;
+        }
+    }
+    if (node->type == AST::AST_IDENTIFIER) {
```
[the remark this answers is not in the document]
mandrys: I reworked the changes and there are no more that many levels of `if`.

systemverilog-plugin/UhdmAst.cc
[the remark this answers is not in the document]
mandrys: This was removed in the recent changes to the MR.

systemverilog-plugin/UhdmAst.cc
mandrys: This part was changed in the recent update to this MR.

systemverilog-plugin/UhdmAst.cc
mandrys: Changes included in the recent update to this MR.

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc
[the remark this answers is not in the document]
wsipak: Thanks a lot!

systemverilog-plugin/UhdmAst.cc
```diff
@@ -399,63 +400,106 @@ static void resolve_wiretype(AST::AstNode *wire_node)
...
-                    force_convert = true;
-                }
-                // Access to slice of memory
-                if (node->children.size() == 1 && node->children[0]->children.size() != 1) {
-                    force_convert = true;
+        return;
+    }
+
+    for (auto *child : node->children) {
+        check_memories(child, scope, memories);
```
kamilrakoczy, line 423:
```suggestion
        check_memories(child, node->type == AST::AST_GENBLOCK ?  scope + "." + node->str : scope, memories);
```
This should allow to remove above ``if``.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -399,63 +400,106 @@ static void resolve_wiretype(AST::AstNode *wire_node)
...
+
+    if (node->str == "\\$readmemh") {
+        if (node->children.size() != 2 || node->children[1]->str.empty() || node->children[1]->type != AST::AST_IDENTIFIER) {
+            log_error("%s:%d: Wrong usage of '\\$readmemh'\n", node->filename.c_str(), node->location.first_line);
+        }
+        std::string name = scope + "." + node->children[1]->str;
+        const auto iter = memories.find(name);
+        if (iter != memories.end()) {
+            add_force_convert_attribute(iter->second, 0);
+        }
```
kamilrakoczy, line 434: I think ``readmemh`` can also be used in another scope than memory is defined, so ideally we should also look in scopes below while looking for memory, but I'm fine with adding TODO comment for now.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2,6 +2,7 @@
 #include <cstring>
 #include <functional>
 #include <limits>
+#include <list>
```
kamilrakoczy, line 5: Remove, no longer needed.

kamilrakoczy (commented):
Minor suggestions, overall LGTM.

wsipak (approved):

systemverilog-plugin/UhdmAst.cc
mandrys: You're right, changed to be this way.

systemverilog-plugin/UhdmAst.cc
mandrys: Added TODO for this modification.

systemverilog-plugin/UhdmAst.cc

mglb (approved):

PR #462, opened 2023-02-23
systemverilog-plugin: fix wires not starting from 0

kamilrakoczy:
Fixes: https://github.com/antmicro/yosys-systemverilog/issues/1508

UHDM-integration-test: https://github.com/chipsalliance/UHDM-integration-tests/pull/713
yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4252472828

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -458,6 +458,16 @@ static void check_memories(AST::AstNode *module_node)
     });
 }
 
+static void warn_start_range(const std::vector<AST::AstNode *> ranges)
+{
+    for (size_t i = 0; i < ranges.size(); i++) {
+        auto start_elem = min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer);
+        if (start_elem != 0) {
+            log_file_warning(ranges[i]->filename, ranges[i]->location.first_line, "Limited support for multirange wires that don't start from 0\n");
+        }
+    }
+}
+
 // This function is workaround missing support for multirange (with n-ranges) packed/unpacked nodes
 // It converts multirange node to single-range node and translates access to this node
 // to correct range
@@ -476,9 +486,6 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
         wire_node->range_valid = true;
         return;
     }
-    size_t size = 1;
-    size_t packed_size = 1;
-    size_t unpacked_size = 1;
     std::vector<AST::AstNode *> ranges;
 
     // Convert only when node is not a memory and at least 1 of the ranges has more than 1 range
@@ -502,11 +509,23 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
         return false;
     }();
     if (convert_node) {
+        // if not already converted
         if (wire_node->multirange_dimensions.empty()) {
-            packed_size = add_multirange_attribute(wire_node, packed_ranges);
-            unpacked_size = add_multirange_attribute(wire_node, unpacked_ranges);
-            size = packed_size * unpacked_size;
-            ranges.push_back(make_range(size - 1, 0));
+            const size_t packed_size = add_multirange_attribute(wire_node, packed_ranges);
+            const size_t unpacked_size = add_multirange_attribute(wire_node, unpacked_ranges);
+            if (packed_ranges.size() == 1 && unpacked_ranges.empty()) {
+                ranges.push_back(packed_ranges[0]->clone());
+            } else if (unpacked_ranges.size() == 1 && packed_ranges.empty()) {
+                ranges.push_back(unpacked_ranges[0]->clone());
+            } else {
+                // currently we have limited support
+                // for multirange wires that doesn't start from 0
+                warn_start_range(packed_ranges);
+                warn_start_range(unpacked_ranges);
+                const size_t size = packed_size * unpacked_size;
+                log_assert(size >= 1);
+                ranges.push_back(make_range(size - 1, 0));
+            }
         }
     } else {
         for (auto r : packed_ranges) {


wsipak (changes requested):
I think it would be very helpful to have a comment for the
`if (wire_node->multirange_dimensions.empty())` block.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -503,10 +500,16 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
...
-            ranges.push_back(make_range(size - 1, 0));
+            size_t packed_size = add_multirange_attribute(wire_node, packed_ranges);
+            size_t unpacked_size = add_multirange_attribute(wire_node, unpacked_ranges);
+            if (packed_ranges.size() == 1 && unpacked_ranges.empty()) {
+                ranges.push_back(packed_ranges[0]->clone());
+            } else if (unpacked_ranges.size() == 1 && packed_ranges.empty()) {
+                ranges.push_back(unpacked_ranges[0]->clone());
+            } else {
+                size_t size = packed_size * unpacked_size;
+                ranges.push_back(make_range(size - 1, 0));
```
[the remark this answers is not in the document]
kamilrakoczy: ``add_multirange_attribute`` is returning actual size of wire, so it should always return at least ``1``, but you are right, this can be problematic if ``size`` is ``0``. I've added assert to make sure it is always positive.

New run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4281712305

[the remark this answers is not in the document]
mglb: @wsipak: the `add_multirange_attribute` does `(max - min) + 1` for every range, and returns 1 when there are no ranges, so 0 is not possible without overflow. https://github.com/chipsalliance/yosys-f4pga-plugins/blob/f5e5cab4550ce18f707ce2e26dea24890ae0732e/systemverilog-plugin/UhdmAst.cc#L239-L240
But the assert is OK, just in case the function is changed.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -502,11 +499,19 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
...
-            packed_size = add_multirange_attribute(wire_node, packed_ranges);
-            unpacked_size = add_multirange_attribute(wire_node, unpacked_ranges);
-            size = packed_size * unpacked_size;
-            ranges.push_back(make_range(size - 1, 0));
+            size_t packed_size = add_multirange_attribute(wire_node, packed_ranges);
+            size_t unpacked_size = add_multirange_attribute(wire_node, unpacked_ranges);
+            if (packed_ranges.size() == 1 && unpacked_ranges.empty()) {
+                ranges.push_back(packed_ranges[0]->clone());
+            } else if (unpacked_ranges.size() == 1 && packed_ranges.empty()) {
+                ranges.push_back(unpacked_ranges[0]->clone());
```
mglb, line 509: I guess this solution is intentionally limited to 1 range. If so, we should at least add TODO/FIXME comment here describing that it works only for cases with one range.
Or maybe a loop that asserts every range starts with 0, if there is more than one.

Example that doesn't work:
```systemverilog
module top (wire_12_21);
  output [1:2] wire_12_21[2:1];
  wire [1:2] wire_12_21[2:1];
  assign wire_12_21[1][1] = 1'b0;
  assign wire_12_21[2][2] = 1'b1;
endmodule
```

mglb (changes requested):

alaindargelas:
Additional test failing formal verification in BSG tests, please  check if this fix handles this test correctly:
module bsg_hash_bank
( i, index_o);
  input [31:0] i;
  output [1:2] index_o;
  wire [1:2] index_o;
  assign index_o[1] = i[1];
  assign index_o[2] = i[0];
endmodule

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: I've added warning, as we have 1 usage of multirange wire that doesn't start from 0 in opentitan.

kamilrakoczy:
> Additional test failing formal verification in BSG tests, please check if this fix handles this test correctly:
>```
>module bsg_hash_bank ( i, index_o);
>input [31:0] i;
>output [1:2] index_o;
>wire [1:2] index_o;
>assign index_o[1] = i[1];
>assign index_o[2] = i[0];
>endmodule
>```

It looks like the same test that is in: https://github.com/antmicro/yosys-systemverilog/issues/1508. This test is passing formal verification. Did you meant to paste some additional (different) test case here?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -458,6 +458,16 @@ static void check_memories(AST::AstNode *module_node)
     });
 }
 
+static void warn_start_range(const std::vector<AST::AstNode *> ranges)
+{
+    for (size_t i = 0; i < ranges.size(); i++) {
+        auto start_elem = min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer);
+        if (start_elem != 0) {
+            log_file_warning(ranges[i]->filename, ranges[i]->location.first_line, "Limited support for multirange wires that doesn't start from 0\n");
```
wsipak, line 466: I think It should be `don't` instead of `doesn't`, right?

wsipak (changes requested):

wsipak (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -502,11 +509,23 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
...
+            const size_t unpacked_size = add_multirange_attribute(wire_node, unpacked_ranges);
+            if (packed_ranges.size() == 1 && unpacked_ranges.empty()) {
+                ranges.push_back(packed_ranges[0]->clone());
+            } else if (unpacked_ranges.size() == 1 && packed_ranges.empty()) {
+                ranges.push_back(unpacked_ranges[0]->clone());
+            } else {
+                // currently we have limited support
+                // for multirange wires that doesn't start from 0
+                warn_start_range(packed_ranges);
+                warn_start_range(unpacked_ranges);
```
mglb, line 524: Shouldn't this be an error? In such cases results are invalid.
I did test only the snippet posted above, so maybe there are cases where the warning is emitted and the result is correct.

mglb (changes requested):

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: There are cases where result is almost correct, e.g.:
```systemverilog
  module dut();
  parameter int unsigned DataWidth         = 32;
  localparam int unsigned NUM_WORDS  = 10;
  logic [NUM_WORDS-1:1][DataWidth-1:0] rf_reg_q;
  
  for (genvar i = 1; i < NUM_WORDS; i++) begin : g_rf_flops
    assign rf_reg_q[i] = i;
  end
  
  endmodule
 ```
 produces:
 ```verilog
 module dut();

  wire [287:0] rf_reg_q;
  assign rf_reg_q[31:0] = 32'd1;
  assign rf_reg_q[63:32] = 32'd2;
  assign rf_reg_q[95:64] = 32'd3;
  assign rf_reg_q[127:96] = 32'd4;
  assign rf_reg_q[159:128] = 32'd5;
  assign rf_reg_q[191:160] = 32'd6;
  assign rf_reg_q[223:192] = 32'd7;
  assign rf_reg_q[255:224] = 32'd8;
  assign rf_reg_q[287:256] = 32'd9;
endmodule
```
and this case is used in ``opentitan``.
Promoting this warning to error would force us to fix usage of multiranges that doesn't start from 0 and this could take some time. I don't think it is worth doing this now as it is not used very often.

WDYT?

mandrys (approved):

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: We decided offline that it is ok to warn for now.

PR #464, opened 2023-03-06
systemverilog-plugin: set range before range_valid

kamilrakoczy:
This PR sets ``range_left``/``range_right`` before setting ``range_valid`` to prevent negative ranges in AST.

Fixes: https://github.com/antmicro/yosys-systemverilog/issues/1506
yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4342540165

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -511,6 +511,8 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
     if (packed_ranges.empty() && unpacked_ranges.empty()) {
         wire_node->attributes.erase(UhdmAst::packed_ranges());
         wire_node->attributes.erase(UhdmAst::unpacked_ranges());
+        wire_node->range_left = 0;
+        wire_node->range_right = 0;
         wire_node->range_valid = true;
         return;
     }


mandrys (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -511,6 +511,8 @@ static void convert_packed_unpacked_range(AST::AstNode *wire_node)
     if (packed_ranges.empty() && unpacked_ranges.empty()) {
         wire_node->attributes.erase(UhdmAst::packed_ranges());
         wire_node->attributes.erase(UhdmAst::unpacked_ranges());
+        wire_node->range_left = 0;
+        wire_node->range_right = 0;
         wire_node->range_valid = true;
```
mglb, line 516: This results in `range=[0:0]` instead of `range=[-1:0]`, so this isn't really a fix.
```
AST_WIRE <misc.sv:14.42-14.45> [0x5614769a7de0] str='\N11' logic range=[0:0]
```
The condition to get rid of "range" from `dumpAst` output: https://github.com/YosysHQ/yosys/blob/7c5ae560a838db3f35dec0760df6cd8ebb590b80/frontends/ast/ast.cc#L333-L334

The code should rather be:
```suggestion
        wire_node->range_left = -1;
        wire_node->range_right = 0;
        wire_node->range_valid = false;
```

mglb (changes requested):
Note that even with the suggested change this PR still does not fix all AST differences for the snippet. Node order is still different, `AST_RANGE` nodes itself have a range, plugin's wires are `logic`, ports have multiranges. This of course does not prevent the PR merge, but I'm not sure whether https://github.com/antmicro/yosys-systemverilog/issues/1506 should be marked as fixed.

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: We don't want to match yosys AST 100%, especially before simplification. 
Surelog provides us additional information about design (e.g. elaborates it). Surelog also groups some nodes so, we don't want to produce same order of nodes as they appeared in the code. 

``logic`` wires and ``multiranges`` are separate problems.

I've asked Alain what exactly he meant by ``nets are swapped``. If it is just different order of nodes in plugin AST, I believe this PR fixes original issue: https://github.com/antmicro/yosys-systemverilog/issues/1506#issuecomment-1457707886

mglb: My bad, somehow I missed the `[0:0]` in `read_verilog` output after simplification.

mglb (approved):

PR #466, opened 2023-03-08
systemverilog-plugin: update module parameters processing

kamilrakoczy:
This PR updates how we are processing parameters to match it with yosys. It also removes some of no-longer-needed code.

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4355510639
UHDM-integration-test: https://github.com/chipsalliance/UHDM-integration-tests/pull/717

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1970,16 +1970,12 @@ void UhdmAst::process_module()
         current_node->children.insert(current_node->children.begin(), typeNode);
         auto old_top = shared.current_top_node;
         shared.current_top_node = module_node;
-        visit_one_to_many({vpiVariables, vpiNet, vpiArrayNet}, obj_h, [&](AST::AstNode *node) {
-            if (node) {
-                add_or_replace_child(module_node, node);
-            }
-        });
-        visit_one_to_many({vpiInterface, vpiModule, vpiPort, vpiGenScopeArray, vpiContAssign, vpiTaskFunc}, obj_h, [&](AST::AstNode *node) {
-            if (node) {
-                add_or_replace_child(module_node, node);
-            }
-        });
+        visit_one_to_many({vpiVariables, vpiNet, vpiArrayNet, vpiInterface, vpiModule, vpiPort, vpiGenScopeArray, vpiContAssign, vpiTaskFunc}, obj_h,
+                          [&](AST::AstNode *node) {
+                              if (node) {
+                                  add_or_replace_child(module_node, node);
+                              }
+                          });
         make_cell(obj_h, current_node, module_node);
         shared.current_top_node = old_top;
     }
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1919,25 +1919,25 @@ void UhdmAst::process_module()
         });
         std::string module_name = !parameters.empty() ? AST::derived_module_name(type, parameters).c_str() : type;
         auto module_node = shared.top_nodes[module_name];
-        auto cell_instance = vpi_get(vpiCellInstance, obj_h);
+        // true, when Surelog don't have definition of module
+        // if so, we are left setting module parameters to yosys
+        // and don't rename module
+        bool isPrimitive = false;
         if (!module_node) {
             module_node = shared.top_nodes[type];
             if (!module_node) {
                 module_node = new AST::AstNode(AST::AST_MODULE);
                 module_node->str = type;
                 module_node->attributes[UhdmAst::partial()] = AST::AstNode::mkconst_int(2, false, 1);
-                cell_instance = 1;
-                module_name = type;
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -1902,33 +1902,22 @@ void UhdmAst::process_module()
...
-        std::string module_parameters;
+        std::vector<std::pair<RTLIL::IdString, RTLIL::Const>> parameters;
         visit_one_to_many({vpiParamAssign}, obj_h, [&](AST::AstNode *node) {
             if (node && node->type == AST::AST_PARAMETER) {
                 if (node->children[0]->type != AST::AST_CONSTANT) {
                     if (shared.top_nodes.count(type)) {
                         simplify_parameter(node, shared.top_nodes[type]);
-                        log_assert(node->children[0]->type == AST::AST_CONSTANT || node->children[0]->type == AST::AST_REALVALUE);
+                    } else {
+                        simplify_parameter(node, nullptr);
```
mandrys, line 1912: What do we expect from parameter simplification in case where there is no other node of the same type in `shared.top_nodes`?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1957,39 +1946,14 @@ void UhdmAst::process_module()
                         log_assert(node->children[0]->type == AST::AST_CONSTANT || node->children[0]->type == AST::AST_REALVALUE);
                     }
                 }
-                auto parent_node = std::find_if(module_node->children.begin(), module_node->children.end(), [&](AST::AstNode *child) -> bool {
-                    return ((child->type == AST::AST_PARAMETER) || (child->type == AST::AST_LOCALPARAM)) && child->str == node->str &&
-                           // skip real parameters as they are currently not working: https://github.com/alainmarcel/Surelog/issues/1035
-                           child->type != AST::AST_REALVALUE;
-                });
```
mandrys, line 1964: Why don't we want to do this part anymore?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1902,33 +1902,22 @@ void UhdmAst::process_module()
     } else {
         // Not a top module, create instance
         current_node = make_ast_node(AST::AST_CELL);
-        std::string module_parameters;
+        std::vector<std::pair<RTLIL::IdString, RTLIL::Const>> parameters;
         visit_one_to_many({vpiParamAssign}, obj_h, [&](AST::AstNode *node) {
             if (node && node->type == AST::AST_PARAMETER) {
                 if (node->children[0]->type != AST::AST_CONSTANT) {
```
wsipak, line 1908: Shouldn't we assert `node->children.size() > 0` before accessing `children[0]` ?

kamilrakoczy: Right, added.

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: Sometimes we don't need access to cell instance to simplify parameter, e.g. in case of primitive cell and parameter that uses concat. We are expecting ``simplify_parameter`` to reduce this concat to constant value.

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: This part were selecting all parameters that had real value in parameter and was forcing to use ``AST_PARASET`` instead of replacing actual value of parameter as surelog used to report wrong values for real type. This is no longer the case.

I've updated test case to showcase this: https://github.com/chipsalliance/UHDM-integration-tests/pull/717/files#diff-3a52e89ebb9bcd325d69c85bf56ee916ac4dc9990e60241e52d5bd0b14ef89afR20

mandrys: Got it, thanks!

systemverilog-plugin/UhdmAst.cc
mandrys: I see, thanks!

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1902,53 +1902,43 @@ void UhdmAst::process_module()
...
         });
-        // rename module in same way yosys do
-        std::string module_name;
-        if (module_parameters.size() > 60)
-            module_name = "$paramod$" + sha1(module_parameters) + type;
-        else if (!module_parameters.empty())
-            module_name = "$paramod" + type + module_parameters;
-        else
-            module_name = type;
+        std::string module_name = !parameters.empty() ? AST::derived_module_name(type, parameters).c_str() : type;
```
mandrys, line 1921: Could we have some comment here to explain, when and why we need the module to be renamed?

wsipak (approved):

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: Sure, added.

mandrys (approved):

mglb (approved):

PR #468, opened 2023-03-14
Reduce expressions in logic typespecs

wsipak:
Consider this example:
```
  localparam int StateWidth = 9;
  typedef enum logic [StateWidth-1:0] {
    StActive = 9'b100100001
  } state_e;
```

As reported in https://github.com/chipsalliance/Surelog/issues/3499 , Surelog doesn't evaluate the range in the definition of the enum.
The UHDM for the logic typespec of the enum we're getting here is:
```
logic_typespec: , line:5:16, endln:5:38
 |vpiRange:
 \_range: , line:5:22, endln:5:38
   |vpiLeftRange:
   \_operation: , line:5:23, endln:5:35
     |vpiParent:
     \_range: , line:5:22, endln:5:38
     |vpiOpType:11
     |vpiOperand:
     \_ref_obj: (StateWidth), line:5:23, endln:5:33
       |vpiParent:
       \_operation: , line:5:23, endln:5:35
       |vpiName:StateWidth
     |vpiOperand:
     \_constant: , line:5:34, endln:5:35
       |vpiParent:
       \_operation: , line:5:23, endln:5:35
       |vpiDecompile:1
       |vpiSize:64
       |UINT:1
       |vpiConstType:9
   |vpiRightRange:
   \_constant: , line:5:36, endln:5:37
     |vpiParent:
     \_range: , line:5:22, endln:5:38
     |vpiDecompile:0
     |vpiSize:64
     |UINT:0
     |vpiConstType:9
```

We can have this simplified using the `reduceExpr` function, as suggested in the linked Surelog issue.
As a result, what needs to be processed in the SV plugin looks like this:
```
logic_typespec: , line:5:16, endln:5:38
 |vpiRange:
 \_range: , line:5:22, endln:5:38
   |vpiLeftRange:
   \_constant:
     |vpiDecompile:8
     |vpiSize:64
     |INT:8
     |vpiConstType:7
   |vpiRightRange:
   \_constant: , line:5:36, endln:5:37
     |vpiParent:
     \_range: , line:5:22, endln:5:38
     |vpiDecompile:0
     |vpiSize:64
     |UINT:0
     |vpiConstType:9
```

See tests running here: https://github.com/antmicro/yosys-systemverilog/actions/runs/4427522808
This is needed for https://github.com/chipsalliance/yosys-f4pga-plugins/pull/454 .

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -11,6 +11,7 @@
 #include "libs/sha1/sha1.h"
 
 // UHDM
+#include <uhdm/ExprEval.h>
 #include <uhdm/uhdm.h>
 #include <uhdm/vpi_user.h>
 
@@ -2173,6 +2174,19 @@ void UhdmAst::process_typespec_member()
     add_multirange_wire(current_node, packed_ranges, unpacked_ranges);
 }
 
+static UHDM::expr *reduce_expression(const UHDM::any *expr, const UHDM::any *inst, const UHDM::any *pexpr)
+{
+    log_assert(expr);
+    log_assert(inst);
+    log_assert(pexpr);
+
+    bool invalidvalue = false;
+    UHDM::ExprEval eval;
+    UHDM::expr *resolved_operation = eval.reduceExpr(expr, invalidvalue, inst, pexpr);
+    log_assert(!invalidvalue);
+    return resolved_operation;
+}
+
 void UhdmAst::process_enum_typespec()
 {
     // BaseTypespec specifies underlying type of the enum.
@@ -2180,6 +2194,42 @@ void UhdmAst::process_enum_typespec()
     // When base type is not specified in SystemVerilog code, it is assumed to be an int.
     // Type of enum items (constants) is the same as the enum type.
     current_node = make_ast_node(AST::AST_ENUM);
+
+    const uhdm_handle *const handle = (const uhdm_handle *)obj_h;
+    const auto *enum_object = (const UHDM::enum_typespec *)handle->object;
+    const auto *typespec = enum_object->Base_typespec();
+
+    if (typespec && typespec->UhdmType() == UHDM::uhdmlogic_typespec) {
+        // If it's a logic_typespec, try to reduce expressions inside of it.
+        // The `reduceExpr` function needs the whole context of the enum typespec
+        //   so it's called here instead of `process_operation` or any other more specific function.
+
+        const UHDM::logic_typespec *logic_typespec_obj = enum_object->Base_typespec()->Cast<const UHDM::logic_typespec *>();
+        std::vector<UHDM::range *> ranges;
+        // Check if ranges exist, as Ranges() returns a pointer to std::vector.
+        if (logic_typespec_obj->Ranges()) {
+            ranges = *(logic_typespec_obj->Ranges());
+        }
+        for (UHDM::range *range_obj : ranges) {
+            // For each range, take both left and right and reduce them if they're of type uhdmoperation.
+            const auto *leftrange_obj = range_obj->Left_expr();
+            const auto *rightrange_obj = range_obj->Right_expr();
+            log_assert(leftrange_obj);
+            log_assert(rightrange_obj);
+
+            if (leftrange_obj->UhdmType() == UHDM::uhdmoperation) {
+                // Substitute the previous leftrange with the resolved operation result.
+                range_obj->Left_expr(reduce_expression(leftrange_obj, enum_object->Instance() ? enum_object->Instance() : enum_object->VpiParent(),
+                                                       enum_object->VpiParent()));
+            }
+            if (rightrange_obj->UhdmType() == UHDM::uhdmoperation) {
+                // Substitute the previous rightrange with the resolved operation result.
+                range_obj->Right_expr(reduce_expression(rightrange_obj, enum_object->Instance() ? enum_object->Instance() : enum_object->VpiParent(),
+                                                        enum_object->VpiParent()));
+            }
+        }
+    }
+
     bool has_base_type = false;
     visit_one_to_one({vpiBaseTypespec}, obj_h, [&](AST::AstNode *node) {
         has_base_type = true;


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
     add_multirange_wire(current_node, packed_ranges, unpacked_ranges);
 }
 
-void UhdmAst::process_enum_typespec()
+void UhdmAst::process_enum_typespec(const UHDM::BaseClass *object)
```
kamilrakoczy, line 2177: I think you should be able to get ``UHDM::BaseClass`` from ``obj_h`` that is already available in this function using:
```
    const uhdm_handle *const handle = (const uhdm_handle *)obj_h;
    const UHDM::BaseClass *const object = (const UHDM::BaseClass *)handle->object;
```

Have you tried this?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
     // BaseTypespec specifies underlying type of the enum.
     // The BaseTypespec has at most one explicit packed dimension (range).
     // When base type is not specified in SystemVerilog code, it is assumed to be an int.
     // Type of enum items (constants) is the same as the enum type.
     current_node = make_ast_node(AST::AST_ENUM);
+
+    const auto *enum_object = object->Cast<const UHDM::enum_typespec *>();
+    const auto *typespec = enum_object->Base_typespec();
+
+    if (typespec && typespec->UhdmType() == UHDM::uhdmlogic_typespec) {
```
kamilrakoczy, line 2188: We will probably want to reduce a lot of different typespec, so maybe it would be better to use ``switch`` instead of ``if``? WDYT?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
+        // If it's a logic_typespec, try to reduce expressions inside of it.
+
+        const auto *logic_typespec_obj = enum_object->Base_typespec()->Cast<const UHDM::logic_typespec *>();
```
kamilrakoczy, line 2191: Maybe lets not use auto if we know exactly what type we are expecting?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
+        std::vector<UHDM::range *> ranges;
+        // Check if ranges exist.
+        if (logic_typespec_obj->Ranges()) {
+            ranges = logic_typespec_obj->Ranges()[0];
```
kamilrakoczy, line 2195: This is a little bit confusing, why do we need to select ``[0]`` element? Please add comment what is internal structure of ``logic_typespec_obj->Ranges()``.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
+        for (UHDM::range *range_obj : ranges) {
+            // For each range, take both left and right and reduce them if they're of type uhdmoperation.
+            const auto *leftrange_obj = range_obj->Left_expr();
+            const auto *rightrange_obj = range_obj->Right_expr();
+            log_assert(leftrange_obj);
+            log_assert(rightrange_obj);
+
+            if (leftrange_obj->UhdmType() == UHDM::uhdmoperation) {
+
+                const auto *operation_obj = leftrange_obj->Cast<const UHDM::operation *>();
```
kamilrakoczy, line 2206: Maybe lets not use auto if we know exactly what type we are expecting?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
+
+                bool invalidvalue = false;
+                UHDM::ExprEval eval;
+                UHDM::expr *resolved_operation =
+                  eval.reduceExpr(operation_obj, invalidvalue, enum_object->Instance() ? enum_object->Instance() : enum_object->VpiParent(),
```
kamilrakoczy, line 2211: What if both ``enum_object->Instance()`` and ``enum_object->VpiParent()`` are nullptr? Is this possible? Maybe we should assert this?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
+            ranges = logic_typespec_obj->Ranges()[0];
+        }
+        for (UHDM::range *range_obj : ranges) {
+            // For each range, take both left and right and reduce them if they're of type uhdmoperation.
+            const auto *leftrange_obj = range_obj->Left_expr();
+            const auto *rightrange_obj = range_obj->Right_expr();
+            log_assert(leftrange_obj);
+            log_assert(rightrange_obj);
+
+            if (leftrange_obj->UhdmType() == UHDM::uhdmoperation) {
```
kamilrakoczy, line 2204: Lets extract this to separate function so we won't need to repeat same ``if`` twice.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,64 @@ void UhdmAst::process_typespec_member()
...
+
+                bool invalidvalue = false;
+                UHDM::ExprEval eval;
+                UHDM::expr *resolved_operation =
+                  eval.reduceExpr(operation_obj, invalidvalue, enum_object->Instance() ? enum_object->Instance() : enum_object->VpiParent(),
+                                  enum_object->VpiParent());
+
+                log_assert(!invalidvalue);
+                // Substitute the previous leftrange with the resolved operation result.
+                range_obj->Left_expr(resolved_operation);
```
kamilrakoczy, line 2216: Do we need to delete previous expression (that we are replacing here) or it is done automatically?

kamilrakoczy (commented):
Generally LGTM, some minor comments and questions.

systemverilog-plugin/UhdmAst.cc
wsipak: At first I didn't realize it could be better not to pass a parameter to the function, so I didn't use `obj_h`, but now I've modified this like you suggested and it works indeed 👍🏻.

systemverilog-plugin/UhdmAst.cc
wsipak: For now I was going to cover that case of Enums specifically, to cover a case in OpenTitan.

However, it seems to me like the function `reduceExpr` is so generic (it takes `UHDM::any *`) that we could potentially have other typespecs handled the same way, so we could just add more conditions to the `if`.

I wasn't targeting any other typespecs in this PR, so if it's not a problem, I would suggest converting it to `switch` or expanding the `if` statement later with other test cases accordingly.

systemverilog-plugin/UhdmAst.cc
wsipak: I don't have a strong preference towards `auto` but quite often I'm advised to use `auto` whenever it fits, so I used it.

I've modified this to specify the type explicitly, I'm not quite sure If we need it though.

systemverilog-plugin/UhdmAst.cc
wsipak: Now I realise it could be confusing indeed, as I used it to dereference a pointer.
I've modified it to use the asterisk dereference operator, and expanded the comment explaining that `Ranges()` returns a pointer.

systemverilog-plugin/UhdmAst.cc
wsipak: AFAIK this shouldn't ever happen. I've added `log_assert` for these pointers.

systemverilog-plugin/UhdmAst.cc
wsipak: I've browsed through the usages of this function in Surelog, and it appears that the old object is never deleted explicitly, and quite often it's just overwritten with the new one, so I expect this shouldn't be an issue.

systemverilog-plugin/UhdmAst.cc
wsipak: I've came up with 2 ideas:

1. We can have a wrapper to `reduceExpr` that will take generic objects and reduce them.
This won't allow to reduce the two `if` statements into one,
as left and right expressions are handled by different methods (`Left_expr` and `Right_expr`),
but if we handle more cases later, we'll be able to use the same function.

2. We can have a wrapper to `reduceExpr` that will specifically take a `UHDM::range *`.
With this, we can replace the two `if` statements with a single call to the function.
This function would be useful only for `UHDM::range *` so it won't be as generic as 1).

I've implemented 1) — I've extracted some of the code to a function but there are still two `if` statements.
I'm not really against 2) but I thought 1) would be better.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,60 @@ void UhdmAst::process_typespec_member()
...
+            log_assert(leftrange_obj);
+            log_assert(rightrange_obj);
+
+            if (leftrange_obj->UhdmType() == UHDM::uhdmoperation) {
+                // Substitute the previous leftrange with the resolved operation result.
+                range_obj->Left_expr(reduce_expression(leftrange_obj, enum_object->Instance() ? enum_object->Instance() : enum_object->VpiParent(),
+                                                       enum_object->VpiParent()));
+            }
+            if (rightrange_obj->UhdmType() == UHDM::uhdmoperation) {
+                // Substitute the previous leftrange with the resolved operation result.
```
kamilrakoczy, line 2224:
```suggestion
                // Substitute the previous rightrange with the resolved operation result.
```

kamilrakoczy (approved):

mandrys (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2173,13 +2174,60 @@ void UhdmAst::process_typespec_member()
...
+                                                       enum_object->VpiParent()));
+            }
+            if (rightrange_obj->UhdmType() == UHDM::uhdmoperation) {
+                // Substitute the previous rightrange with the resolved operation result.
+                range_obj->Right_expr(reduce_expression(rightrange_obj, enum_object->Instance() ? enum_object->Instance() : enum_object->VpiParent(),
+                                                        enum_object->VpiParent()));
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
wsipak:
```
/* Tries to reduce any expression into a constant, returns the orignal
     expression if fails. If an invalid value is found in the process,
     invalidValue will be set to true */
  expr* reduceExpr(const any* object, bool& invalidValue, const any* inst,
                   const any* pexpr, bool muteErrors = false);
```

The `reduceExpr` function needs to be provided with the `inst` and `pexpr` pointers
which as I assume are supposed to be the instance and parent object.

The function works when it's executed from within the context of the Enum Typespec (`->Instance()` and `->VpiParent()` of an `enum_typespec` object).
The function doesn't work when provided with a narrower context (for example with the operation that was extracted from the `enum_typespec`.

So, if we wanted to execute it from within `process_logic_typespec`, `process_range` or anything narrower, we would need to provide the Enum typespec anyway.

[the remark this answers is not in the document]
mglb: Oh. Okay, too much work for now. Please just add a comment (probably `TODO`?) with this explanation. Can be shortened a bit to just state that we would need to pass context-objects to reduceExpr.

mglb (approved):
Please just add a comment then merge. Thanks.

PR #469, opened 2023-03-17
systemverilog-plugin: add parameter type propagation through hierarchy

joennlae:
Hello :-) First of all really cool project 💯

We are working on parsing a fully-fledged SoC through yosys and further.

This feature is something we need, but there will be more that we will be adding over the upcoming weeks.

All our IPs and RTL are open-source here: https://github.com/pulp-platform/iguana.

We are currently working on `baby_iguana`. And work our way up from there. A pickle file can be found here: 

https://gist.github.com/joennlae/35baf335b0124caf75869ab3e5edce33

The feature still needs to be fully completed. For example, when propagating types, the input with a custom type is an array.

Here:
https://gist.github.com/joennlae/35baf335b0124caf75869ab3e5edce33#file-baby_iguana-pickle-sv-L166

```sv
    // .... this is passed two hierarchies above in module dm_csrs
    .dtype            ( logic [$bits(dmi_resp_o)-1:0] ), // logic [33:0]
    // .... in fifo_v3
    // actual memory
    dtype [FifoDepth - 1:0] mem_n, mem_q;
```
And the replacement in yosys via the uhdm bridge is:
```v
  (* src = "/home/janniss/Documents/iguana/build/baby_iguana_top.pickle.sv:166.36-166.41" *)
  reg [33:0] mem_q;
```

Even though it should be
```v
  (* src = "/home/janniss/Documents/iguana/build/baby_iguana_top.pickle.sv:166.36-166.41" *)
  reg [67:0] mem_q;
```

The range information is overwritten.

But we will be working on it. I just wanted to open the feedback cycle very early, and I look forward to a fruitful cooperation :-)


joennlae:
I ran the `clang-format` :-).

wsipak:
Thank you for the PR. Your contribution is very much appreciated.

You might want to know that we have on-going work towards supporting type parameters here: https://github.com/chipsalliance/yosys-f4pga-plugins/pull/454

The linked PR targets slightly different cases, though, so it's not redundant.

Ideally we would like to unify the code parts that serve the same purpose for all type parameters, for example the concept of renaming modules with the names and values of parameters, and then merge parts that are supposed to handle specific variations of type parameters syntax, like the one you’re presenting here (`.dtype ( logic [$bits(dmi_resp_o)-1:0] )
`).

It would be very helpful to have a minimal code snippet for the type parameter, without the need to reference the whole `baby_iguana` file.
Usually we create one-file tests to demonstrate a specific syntax. You can find a bunch of examples here: https://github.com/chipsalliance/UHDM-integration-tests/tree/master/tests
And if you're interested in the `type parameter` syntaxes we were targeting in the first place, you can see this:
https://github.com/chipsalliance/UHDM-integration-tests/pull/707/files

PR #470, opened 2023-03-20
systemverilog-plugin: add support for custom type with multiranges inside struct/union

kamilrakoczy:
Fixes: https://github.com/chipsalliance/yosys-f4pga-plugins/issues/467

This PR adds support for custom type with multirange inside struct/union. Hierarchical (using ``dot``) access is still restricted to single range.

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4477563172
UHDM-integration-tests: https://github.com/chipsalliance/UHDM-integration-tests/pull/718

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -619,7 +619,8 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
         if (c->type == AST::AST_RANGE) {
             // Currently supporting only 1 range
             if (struct_range) {
-                log_error("Currently support for dot-access is limited to single range\n");
+                log_file_error(struct_range->filename, struct_range->location.first_line,
+                               "Currently support for dot-access is limited to single range\n");
             }
             struct_range = c;
         }
@@ -636,7 +637,8 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
         // with multirange
         // Currently support only special access to 2 dimensional packed element
         // when selecting single range
-        if (struct_range && current_struct_elem->multirange_swapped.size() == 2 /* 2 dimensional */) {
+        log_assert(current_struct_elem->multirange_dimensions.size() % 2 == 0);
+        if (struct_range && (current_struct_elem->multirange_dimensions.size() / 2) == 2) {
             // get element size in number of bits
             const int single_elem_size = current_struct_elem->children.front()->range_left + 1;
             left = AST::AstNode::mkconst_int(single_elem_size * current_struct_elem->multirange_dimensions.back(), true);
@@ -847,9 +849,10 @@ static int simplify_struct(AST::AstNode *snode, int base_offset, AST::AstNode *p
             // embedded struct or union
             width = simplify_struct(node, base_offset + offset, parent_node);
             if (!node->multirange_dimensions.empty()) {
-                // reverse-iterate over multiranges, skip right_range
-                for (auto rit = node->multirange_dimensions.rbegin(); rit != node->multirange_dimensions.rend(); rit += 2) {
-                    width *= *rit;
+                // Multiply widths of all dimensions.
+                // `multirange_dimensions` stores (repeating) pairs of [offset, width].
+                for (size_t i = 1; i < node->multirange_dimensions.size(); i += 2) {
+                    width *= node->multirange_dimensions[i];
                 }
             }
             // set range of struct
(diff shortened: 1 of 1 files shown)


wsipak:
I'm not quite sure if this belongs to the scope of this PR; However, it seems that here:
https://github.com/antmicro/yosys-f4pga-plugins/blob/kr/union_custom_type_multirange/systemverilog-plugin/UhdmAst.cc#L675
The object pointed by `right` is cloned, and the old pointer is overwritten, which means there's a memory leak.
That specific line isn't modified here, but It's related so I thought it was worth mentioning.

wsipak (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -834,9 +846,10 @@ static int simplify_struct(AST::AstNode *snode, int base_offset, AST::AstNode *p
             // embedded struct or union
             width = simplify_struct(node, base_offset + offset, parent_node);
             if (!node->multirange_dimensions.empty()) {
-                int number_of_structs = 1;
-                number_of_structs = node->multirange_dimensions.back();
-                width *= number_of_structs;
+                // reverse-iterate over multiranges, skip right_range
+                for (auto rit = node->multirange_dimensions.rbegin(); rit != node->multirange_dimensions.rend(); rit += 2) {
+                    width *= *rit;
+                }
```
mglb, line 852: "skip right_range" → "skip offset"?
I think reverse iteration here is a bit too clever, doubly when emphasized in the comment.
```suggestion
                // Multiply widths of all dimensions.
                // `multirange_dimensions` stores (repeating) pairs of [offset, width].
                for (size_t i = 1; i < node->multirange_dimensions.size(); i += 2) {
                    width *= node->multirange_dimensions[i];
                }
```

mglb (changes requested):
Looks OK apart from two small comments.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -635,10 +617,44 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
...
+        left = AST::AstNode::mkconst_int(current_struct_elem->range_left, true);
+        right = AST::AstNode::mkconst_int(current_struct_elem->range_right, true);
+        break;
+    case AST::AST_STRUCT:
+    case AST::AST_UNION:
+        // TODO(krak): add proper support for accessing struct/union elements
+        // with multirange
+        // Currently support only special access to 2 dimensional packed element
+        // when selecting single range
+        if (struct_range && current_struct_elem->multirange_swapped.size() == 2 /* 2 dimensional */) {
```
mglb, line 638: Wait, shouldn't you take into account `multirange_swapped` values somewhere below? At least in an assert.

kamilrakoczy: Oh, I've assumed that size of ``multirange_swapped`` equals to half of size of ``multirange_dimensions`` (and it should always be the case), but you are right, this can be confusing. I've changed this to just ``multirange_dimensions.size() / 2``.

kamilrakoczy:
New yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4540552775

systemverilog-plugin/UhdmAst.cc
mglb: It does, I think?
As for readability, I'm not sure this division + assert with modulo is much better. I think `const int dimensions_num = multirange_swapped.size()` and then `dimensions_num == 2` in `if` would be best.

However, what I had in mind in the first comment is:
shouldn't we take into account values stored in `multirange_swapped` when using values like `current_struct_elem->children.front()->range_left` in calculations? I know `multirange_swapped` tells whether ranges are swapped, but I don't know how it affects this use case (if at all).

mandrys (approved):

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: > It does, I think? As for readability, I'm not sure this division + assert with modulo is much better. I think `const int dimensions_num = multirange_swapped.size()` and then `dimensions_num == 2` in `if` would be best.

At first I tried this, but compiler errors-out with message about cross initialization.

> However, what I had in mind in the first comment is: shouldn't we take into account values stored in `multirange_swapped` when using values like `current_struct_elem->children.front()->range_left` in calculations? I know `multirange_swapped` tells whether ranges are swapped, but I don't know how it affects this use case (if at all).

I don't think we need to take into account value stored in ``multirange_swapped`` in the code that I'm changing.

Correct order of ranges (right range lower than left range) is assured by ``add_multirange_attribute`` function and I'm only adjusting size of single element.

``multirange_swapped`` needs to be taken into account when selecting actual element, but I think this is beyond the scope for this PR.

mglb: Okay, good to know. Thanks.

mglb (approved):

PR #473, opened 2023-03-27
systemverilog-plugin: fix assignments on declaration

kamilrakoczy:
This PR fixes assignments on declaration in more generic way. Assignments on declaration in genfor blocks are handled separately, so they are left intact.

UHDM-integration-tests: https://github.com/chipsalliance/UHDM-integration-tests/pull/719
yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4531860494

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -2636,9 +2636,26 @@ void UhdmAst::process_assignment(const UHDM::BaseClass *object)
 
     visit_one_to_one({vpiLhs, vpiRhs}, obj_h, [&](AST::AstNode *node) {
         if (node) {
-            if (node->type == AST::AST_PARAMETER || node->type == AST::AST_LOCALPARAM) {
+            // fix node types for some assignments
+            // yosys requires that declaration of variable
+            // and assignment are separated
+            switch (node->type) {
+            case AST::AST_WIRE:
+                // wires can be declarated inside initialization block of for block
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2571,9 +2571,26 @@ void UhdmAst::process_assignment(const UHDM::BaseClass *object)
...
+                if (AST::AstNode *for_block = find_ancestor({AST::AST_BLOCK})) {
+                    if (for_block->str.find("$fordecl_block") != std::string::npos)
+                        break;
+                }
+                [[fallthrough]];
+            case AST::AST_PARAMETER:
+            case AST::AST_LOCALPARAM:
                 node->type = AST::AST_IDENTIFIER;
-            }
+                node->attributes.erase(UhdmAst::packed_ranges());
```
wsipak, line 2588: I think we should first `delete` the objects stored in attributes.

wsipak (changes requested):

mglb (approved):

kamilrakoczy:
Previous yosys-systemverilog run discovered, that opentitan synthesis failed with this change. It was due to additional children in ``AST_IDENTIFIER`` that was leftover after transforming it from ``AST_WIRE``. I've fixed this.

new yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4598266847

PR #474, opened 2023-03-28
systemverilog-plugin: Fix IdString initialization and memory leaks of AstNode objects

wsipak:
CI run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4563494012
https://github.com/antmicro/yosys-systemverilog/actions/runs/4552911545

These changes are rebased on the top of https://github.com/chipsalliance/yosys-f4pga-plugins/pull/463 as we are aiming to merge these at once.

Co-authored with @mglb.

Changes:
- The private IdStrings renamed to start with $, as this is convention used in Yosys (see: [verific frontend](https://github.com/YosysHQ/yosys/blob/f35bdaa5278b408e94ac8c97a06c3f7f0fe1773a/frontends/verific/verific.cc#L196), [ID() macro](https://github.com/YosysHQ/yosys/blob/f35bdaa5278b408e94ac8c97a06c3f7f0fe1773a/kernel/yosys.h#L344)).
- IdStrings are initialized all at once on plugin registration and removed after the design is parsed. This is required because, for some reason, Yosys can generate slightly different output depending on how much extra IdStrings are there.
- Memory leaks that would cause some IdStrings to remain undeleted after the plugin does its job. As IdString references are counted in Yosys, it's essential not to leak the objects created by the plugin.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -42,41 +42,44 @@ enum AstNodeTypeExtended {
 }
 } // namespace AST
 
-/*static*/ const IdString &UhdmAst::partial()
+namespace attr_id
 {
-    static const IdString id("\\partial");
-    return id;
-}
-/*static*/ const IdString &UhdmAst::packed_ranges()
-{
-    static const IdString id("\\packed_ranges");
-    return id;
-}
-/*static*/ const IdString &UhdmAst::unpacked_ranges()
-{
-    static const IdString id("\\unpacked_ranges");
-    return id;
-}
-/*static*/ const IdString &UhdmAst::force_convert()
-{
-    static const IdString id("\\force_convert");
-    return id;
-}
-/*static*/ const IdString &UhdmAst::is_imported()
-{
-    static const IdString id("\\is_imported");
-    return id;
-}
-/*static*/ const IdString &UhdmAst::is_simplified_wire()
-{
-    static const IdString id("\\is_simplified_wire");
-    return id;
-}
+static IdString partial;
+static IdString packed_ranges;
+static IdString unpacked_ranges;
+static IdString force_convert;
+static IdString is_imported;
+static IdString is_simplified_wire;
+static IdString low_high_bound;
+}; // namespace attr_id
 
-/*static*/ const IdString &UhdmAst::low_high_bound()
+/*static*/ const IdString &UhdmAst::partial() { return attr_id::partial; }
+/*static*/ const IdString &UhdmAst::packed_ranges() { return attr_id::packed_ranges; }
+/*static*/ const IdString &UhdmAst::unpacked_ranges() { return attr_id::unpacked_ranges; }
+/*static*/ const IdString &UhdmAst::force_convert() { return attr_id::force_convert; }
+/*static*/ const IdString &UhdmAst::is_imported() { return attr_id::is_imported; }
+/*static*/ const IdString &UhdmAst::is_simplified_wire() { return attr_id::is_simplified_wire; }
+/*static*/ const IdString &UhdmAst::low_high_bound() { return attr_id::low_high_bound; }
+
+void UhdmAst::static_init()
 {
-    static const IdString id("\\low_high_bound");
-    return id;
+    // Initialize only once
+    static bool already_initialized = false;
+    if (already_initialized)
+        return;
+    already_initialized = true;
+
+    // Actual initialization
+
+    // Register IdStrings. Can't be done statically, as the IdString class uses resources created during Yosys initialization which happens after
+    // static initialization of the plugin when everything is statically linked.
+    attr_id::partial = IdString("$systemverilog_plugin$partial");
+    attr_id::packed_ranges = IdString("$systemverilog_plugin$packed_ranges");
+    attr_id::unpacked_ranges = IdString("$systemverilog_plugin$unpacked_ranges");
+    attr_id::force_convert = IdString("$systemverilog_plugin$force_convert");
+    attr_id::is_imported = IdString("$systemverilog_plugin$is_imported");
+    attr_id::is_simplified_wire = IdString("$systemverilog_plugin$is_simplified_wire");
+    attr_id::low_high_bound = IdString("$systemverilog_plugin$low_high_bound");
 }
 
 static void sanitize_symbol_name(std::string &name)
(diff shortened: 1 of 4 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -632,6 +699,7 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
             // There should be only 1 AST_DOT node children
             log_assert(!sub_dot);
             sub_dot = expand_dot(current_struct_elem, c);
+            break;
```
kamilrakoczy, line 702: I think we don't want this, multiple ``AST_DOTs`` doesn't make sense, but I think we should check for this.

Beside that, I think after ``AST_DOT`` there can be ``AST_RANGE`` and with this break, we won't find it.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1588,12 +1661,18 @@ static void add_or_replace_child(AST::AstNode *parent, AST::AstNode *child)
             if ((*it)->attributes.count(UhdmAst::packed_ranges()) && child->attributes.count(UhdmAst::packed_ranges())) {
                 if ((!(*it)->attributes[UhdmAst::packed_ranges()]->children.empty() &&
                      child->attributes[UhdmAst::packed_ranges()]->children.empty())) {
+
+                    if (child->attributes.count(UhdmAst::packed_ranges()))
+                        delete child->attributes[UhdmAst::packed_ranges()];
```
kamilrakoczy, line 1666:
```suggestion
                    delete_attribute(current_node, UhdmAst::packed_ranges());
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1588,12 +1661,18 @@ static void add_or_replace_child(AST::AstNode *parent, AST::AstNode *child)
...
                     child->attributes[UhdmAst::packed_ranges()] = (*it)->attributes[UhdmAst::packed_ranges()]->clone();
                 }
             }
             if ((*it)->attributes.count(UhdmAst::unpacked_ranges()) && child->attributes.count(UhdmAst::unpacked_ranges())) {
                 if ((!(*it)->attributes[UhdmAst::unpacked_ranges()]->children.empty() &&
                      child->attributes[UhdmAst::unpacked_ranges()]->children.empty())) {
+
+                    if (child->attributes.count(UhdmAst::unpacked_ranges()))
+                        delete child->attributes[UhdmAst::unpacked_ranges()];
```
kamilrakoczy, line 1675:
```suggestion
                    delete_attribute(current_node, UhdmAst::unpacked_ranges());
```

kamilrakoczy (changes requested):
Generally looks good, but please squash commits and put changes that changes behavior to separate commit, so it would be easier to test it in the future if something fails.

systemverilog-plugin/UhdmAst.cc
wsipak: Applied with `s/current_node/child/`.

systemverilog-plugin/UhdmAst.cc
wsipak: Now I realise there's the `log_assert(!sub_dot)` so the valid pointer won't be overwritten. It looks like the `break` is unneeded, so I've removed it and run the CI again.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -191,20 +250,20 @@ static void visitEachDescendant(AST::AstNode *node, const std::function<void(AST
 static void add_multirange_wire(AST::AstNode *node, std::vector<AST::AstNode *> packed_ranges, std::vector<AST::AstNode *> unpacked_ranges,
                                 bool reverse = true)
 {
+    delete_attribute(node, UhdmAst::packed_ranges());
```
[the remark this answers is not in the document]
mglb: Look one line below. It just assigns new pointer in the place of the old one. The attribute has been already removed (overwritten), this change just deletes the object it has been pointing to, which did leak before.

Maybe the existing logic is wrong here, but it is also definitely wrong from memory management PoV. This change fixes memory management issue. The logic, if needed, should be investigated independently.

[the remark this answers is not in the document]
kamilrakoczy: Oh, right, so it doesn't change behavior.

kamilrakoczy (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -42,41 +43,99 @@ enum AstNodeTypeExtended {
...
+    attr_id::partial = IdString();
 }
-/*static*/ const IdString &UhdmAst::is_imported()
+
+static void delete_attribute(AST::AstNode *node, const IdString &attribute)
 {
-    static const IdString id("\\is_imported");
-    return id;
+    // Delete the selected attribute if it exists.
+    // Does nothing if the node doesn't exist, or the attribute doesn't exists.
```
mglb, line 102: Put comment above definition.
```
// Delete the selected attribute if it exists.
// Does nothing if the node doesn't exist, or the attribute doesn't exists.
static void delete_attribute(AST::AstNode *node, const IdString &attribute)
{
```
(no "suggestion" snippet because github hates us.)

systemverilog-plugin/UhdmAst.cc
```diff
@@ -42,41 +43,99 @@ enum AstNodeTypeExtended {
...
 }
-/*static*/ const IdString &UhdmAst::is_simplified_wire()
+
+static void delete_internal_attributes(AST::AstNode *node)
 {
-    static const IdString id("\\is_simplified_wire");
-    return id;
+    // Delete all attributes that belong to the SV plugin.
+    // The attributes beloning to Yosys are *not* deleted here.
+
```
mglb, line 116:
```
// Delete all attributes that belong to the SV plugin.
// The attributes beloning to Yosys are *not* deleted here.
static void delete_internal_attributes(AST::AstNode *node)
{
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -42,41 +43,99 @@ enum AstNodeTypeExtended {
...
 }
 
-/*static*/ const IdString &UhdmAst::low_high_bound()
+static void delete_children(AST::AstNode *node)
 {
-    static const IdString id("\\low_high_bound");
-    return id;
+    // Delete all children nodes.
+    // Does *not* delete attributes.
+    // This function exists as Yosys's function node->delete_children() does remove all children and attributes.
```
mglb, line 130:
```
// Delete all children nodes. Does *not* delete attributes.
// This function exists as Yosys's function node->delete_children() does remove all children and all attributes.
static void delete_children(AST::AstNode *node)
{
```

mglb (changes requested):

systemverilog-plugin/UhdmAst.cc
wsipak: Alright, done.

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc

mglb (approved):

PR #480, opened 2023-04-06
Support struct with a range in simplify

mandrys:
This PR adds support for simplifying structs and unions with a range.
It fixes a failure of ```Assert 'node->type == AST_STRUCT_ITEM' failed in frontends/ast/simplify.cc:319```, spotted before copying simplify from Yosys to the plugin codebase.

Added test for the fix: https://github.com/chipsalliance/UHDM-integration-tests/pull/721
Test yosys-systemverilog CI run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4629461693

### DIFF
--- a/systemverilog-plugin/third_party/yosys/simplify.cc
+++ b/systemverilog-plugin/third_party/yosys/simplify.cc
@@ -175,6 +175,22 @@ static int size_packed_struct(Yosys::AST::AstNode *snode, int base_offset)
 	bool is_union = (snode->type == Yosys::AST::AST_UNION);
 	int offset = 0;
 	int packed_width = -1;
+
+	// embedded struct or union with range?
+	auto it = std::remove_if(snode->children.begin(), snode->children.end(), [](Yosys::AST::AstNode *node) { return node->type == Yosys::AST::AST_RANGE; });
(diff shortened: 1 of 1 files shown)


kamilrakoczy (commented):
I think we should start moving custom simplification code to this file. Maybe we could move (remove code from custom simplification step and add here) whole struct simplification in this PR?

If it is too much work for now, can we remove part that is moved in this PR from custom simplification step, or it is still needed there?

wsipak (approved):

mandrys:
> I think we should start moving custom simplification code to this file. Maybe we could move (remove code from custom simplification step and add here) whole struct simplification in this PR?
> 
> If it is too much work for now, can we remove part that is moved in this PR from custom simplification step, or it is still needed there?

I've run [tests](https://github.com/antmicro/yosys-systemverilog/actions/runs/4668080473) on the plugin with custom code of struct simplification replaced by the one from _simplify.cc_ and a lot of them started failing. Also, removing only the part that was moved to _simplify.cc_ from the custom struct simplify code caused [test failures](https://github.com/antmicro/yosys-systemverilog/actions/runs/4668623723).
I would leave the PR as it is, with this fix only, and open next PR for moving the custom code to _simplify.cc_.

Issue #432, opened 2023-01-04
Split apart the GitHub Actions to be per plugin


mithro:
There seems to be a single massive GitHub Action which takes ~3 hours to build.

Really there should be a GitHub Action per Yosys plugin.

mithro:
The GitHub Actions should also probably check;
 * Building against Yosys in `hdl/conda-eda`.
 * Building against upstream Yosys HEAD.
 * Building against the latest Yosys stable release.

Issue #431, opened 2023-01-04
Make sure that plugins are being compiled with the same compiler as Yosys


mithro:
If Yosys is compiled with the conda compilers then the plugins should also be compiled with the conda compilers.

PR #485, opened 2023-04-14
Support Mac build, update GHA build to take into account latest Surelog compatibility changes, remove conda

### LINKED ISSUE #74: macOS: There's no `-D` switch in BSD `install`
One of the latest commits (https://github.com/SymbiFlow/yosys-symbiflow-plugins/commit/c613a356f3b5a6e93d2ba64f3568f4e47cef4768) added the `install -D` command.

After that change building the *plugins* on macOS fails in the https://github.com/litex-hub/litex-conda-eda repository.

The BSD's `install -d` seems to have a similar meaning to the GNU's `install -D` but I'm not sure about that:
```
     -d      Create directories.  Missing parent directories are created as required.
```

It'd be great if the script could either have `install` used universally to both implementations or simply have the commands replaced in the commit with `install` as a fail-over for macOS.

### LINKED ISSUE #496: synth_quicklogic name conflict
There is a `synth_quicklogic` pass in [yosys: techlibs/quicklogic](https://github.com/YosysHQ/yosys/blob/master/techlibs/quicklogic/synth_quicklogic.cc) as well  as in [yosys-f4pga-plugins: ql-qlf](https://github.com/chipsalliance/yosys-f4pga-plugins/blob/main/ql-qlf-plugin/synth_quicklogic.cc).

If Yosys has its version compiled in, it is not possible to load the yosys-f4pga-plugin, as a pass with that name is already registered:
```
ERROR: Unable to register pass 'synth_quicklogic', pass already exists!
```

Not sure how to resolve, probably our pass needs to be renamed to not clash with the other name ?

timkpaine:
Opening this for visibility, **please don't "Approve and Run" the CI as I will continue to be noisy**. I am testing actions on my fork.

- Support MacOS - Fixes #74 
- Install Surelog / UHDM in normal manner, remove bespoke conda channel dependency
  - pins Surelog in CI, easily updatable manner
- Aligns with yosys-0.17 to match conda version, sets up for easier upgrading in CI
  - renames `synth_quicklogic` to `synth_quicklogic_f4pga` (fixes https://github.com/chipsalliance/yosys-f4pga-plugins/issues/496)
  - pins version of pmgen (not a fix for, but related to https://github.com/chipsalliance/yosys-f4pga-plugins/issues/526)
- Port from unsupported `json11` to `nlohmann-json` (`nlohmann-json` also used by Surelog as of https://github.com/chipsalliance/Surelog/pull/3827, so this is a good alignment)

This PR has been validated locally against a non-conda Surelog / UHDM and source-installed Yosys on MacOS 13, but is still very much a work-in-progress as I work through lots of CI changes

xref / supercedes: #445
xref: #431 
xref: #432
xref: https://github.com/chipsalliance/yosys-f4pga-plugins/issues/365


timkpaine:
Where I'm at in case curious, still working through some fun Mac issues on CI
Screenshot 2023-04-14 at 19 53 01

timkpaine:
Issues: 
[ ] qlf plugin tests for litex arty fails when `-abc9` passed

timkpaine:
xdc minilitex test passes with `-abc9` when `-g` flag passed to Yosys 🤦

.github/workflows/ci.yml
```diff
@@ -41,40 +41,110 @@ jobs:
...
-    - name: Install
+    - name: Install Ubuntu Dependencies
       run: |
         sudo apt-get update
         sudo apt-get install git g++-9 build-essential bison flex \
           libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot \
           pkg-config libboost-system-dev libboost-python-dev \
-          libboost-filesystem-dev zlib1g-dev clang-format-8 cmake
+          libboost-filesystem-dev zlib1g-dev clang-format-8 cmake \
+          nlohmann-json3-dev iverilog
```
hzeller, line 59: Aren't there googletest dependencies available in Ubuntu and homebrew that we can use directly from the package manager ?

That way we don't need the separate googletest compilation step below.

hzeller:
I think removing the gtest from third_party and then use what is on the system is a good improvement that can be done in a smallish, separate PR and reduces the size of this one.

.github/workflows/ci.yml
timkpaine: I'm not sure if they compile to support c++ 17 string views which are needed, I can test though

hzeller:
Should the workflow be run ?
Did you have time to break out the "don't use gtest from third_party/ but always in system" change ?

timkpaine:
sorry will be getting back to this shortly, been a busy couple months 😢

.github/workflows/ci.yml
```diff
@@ -41,40 +41,110 @@ jobs:
...
+      run: |
+          export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+          set -ex
+
+          python -m pip install -vvv orderedmultidict
+
+          git clone https://github.com/chipsalliance/Surelog.git
+          pushd Surelog
+          git submodule update --init --recursive
+          cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DSURELOG_USE_HOST_GTEST=ON -DPYTHON_EXECUTABLE=`which python` -DPython3_EXECUTABLE=`which python` -S . && cmake --build build && sudo cmake --install build
```
mglb, line 138: Please provide some convenient (i.e. easily visible in PRs when changed) way to specify Surelog revision to use. Perhaps a file containing a hash?
We try to track Surelog master in systemverilog-plugin, however:
- Sometimes (for example now) bugs are discovered in Surelog/UHDM which prevent use of a recent revision in the plugin. We have to wait for fixes in such situations, and use slightly older revision until then.
- By always fetching master branch we'll have non-deterministic CI runs. The same code can first pass CI and then fail on a re-run if there was some merge in Surelog in the meantime.

timkpaine: done, pinned to second-to-latest Surelog due to compatibility

timkpaine:
This PR is ready, please do NOT enable the workflows yet as there appears to be a new (~1-2 days ago) issue with homebrew ccache. 

Given the size of this PR, even though the complexity is quite low, I've started breaking changes out into other PRs:
- https://github.com/chipsalliance/yosys-f4pga-plugins/pull/544
- https://github.com/chipsalliance/yosys-f4pga-plugins/pull/543

There are a few major follow-on issues to investigate:
- this repo pinned yosys to 0.17 via a custom conda channel. We now pull direct from the repo (we could also pull from conda-forge, where yosys is now also available), but I left the version at 0.17 which is quite old now. This should be upgraded to the latest version if possible
- package for various distribution channels (I will do homebrew and conda-forge)

timkpaine:
Everything except the linux ql-qlf plugin is working and passing all tests.

I believe this is due to little differences is path handling, I hit a bunch of those for e.g. how icarus deals with include paths, etc.

All conda reliance is completely removed. 
I also have to rebase after https://github.com/chipsalliance/yosys-f4pga-plugins/pull/546

timkpaine:
closing, cleaning up my backlog

PR #488, opened 2023-04-19
Take name from the AST_DOT node in pattern assignment

mandrys:
As reported in issue https://github.com/chipsalliance/yosys-f4pga-plugins/issues/482, in cases were typedef struct parameter is used, only the value of the first struct field was added to the whole parameter value, and the rest was ignored - it was caused by the fact that the string of the struct fields were skipped when adding the values of struct, so all of them had the same name, hence they were treated as one field.

Test _yosys-systemverilog_ CI run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4743420278
Test for the plugin: https://github.com/chipsalliance/UHDM-integration-tests/pull/723

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3586,13 +3586,22 @@ void UhdmAst::process_assignment_pattern_op()
         std::map<size_t, AST::AstNode *> ordered_children;
         visit_one_to_many({vpiOperand}, obj_h, [&](AST::AstNode *node) {
             if (node->type == AST::AST_ASSIGN || node->type == AST::AST_ASSIGN_EQ || node->type == AST::AST_ASSIGN_LE) {
-                // Find at what position in the concat should we place this node
-                auto key = node->children[0]->str;
-                key = key.substr(key.find('.') + 1);
+                // Get the name of the parameter or it's child, to which the pattern is assigned.
+                std::string key;
+                if (!node->children.empty() && !node->children[0]->children.empty() &&
+                    node->children[0]->children[0]->type == static_cast<AST::AstNodeType>(AST::Extended::AST_DOT)) {
+                    key = node->children[0]->children[0]->str;
+                } else if (!node->children.empty()) {
+                    key = node->children[0]->str;
+                } else {
+                    log_file_error(node->filename, node->location.first_line, "Couldn't find `key` in assignment pattern.\n");
+                }
                 auto param_type = shared.param_types[param_node->str];
                 if (!param_type) {
                     log_error("Couldn't find parameter type for node: %s\n", param_node->str.c_str());
                 }
+                // Place the child node holding the value assigned in the pattern, in the right order,
+                // so the overall value of the param_node is correct.
                 size_t pos =
                   std::find_if(param_type->children.begin(), param_type->children.end(), [key](AST::AstNode *child) { return child->str == key; }) -
                   param_type->children.begin();


systemverilog-plugin/UhdmAst.cc
```diff
@@ -4023,6 +4023,9 @@ void UhdmAst::process_tagged_pattern()
         if (field_name != "default") { // TODO: better support of the default keyword
             auto field = new AST::AstNode(static_cast<AST::AstNodeType>(AST::Extended::AST_DOT));
             field->str = field_name;
+            if (find_ancestor({AST::AST_PARAMETER, AST::AST_LOCALPARAM})) {
+                current_node->children[0]->str = current_node->children[0]->str + "." + field_name;
```
kamilrakoczy, line 4027: I think that instead of adding ``.`` here, we should adapt ``UhdmAst::process_assignment_pattern_op``. It looks like ``UhdmAst::process_assignment_pattern_op`` never got updated to new flow with ``AST_DOT`` and is still looking for ``.`` in string. Instead it should check if node have ``AST_DOT`` as children and take ``key`` as string from it.

https://github.com/chipsalliance/yosys-f4pga-plugins/blob/3a440c4d33e47849e4d974f0bd65bb40272716e9/systemverilog-plugin/UhdmAst.cc#L3590-L3591

Example AST that gets ``UhdmAst::process_assignment_pattern_op``:
```
AST_ASSIGN <dut.sv:0.0-0.0> [0x55dd211d51b0]               
  AST_IDENTIFIER <dut.sv:0.0-0.0> [0x55dd21194140] str='\my_package::my_parameter' 
    AST_DOT <dut.sv:0.0-0.0> [0x55dd21221bb0] str='a'      
  AST_CONSTANT <dut.sv:0.0-0.0> [0x55dd21195210] bits='1'(1) range=[0:0] int=1
```

mandrys:
New test _yosys_systemverilog_ CI run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4752722061

systemverilog-plugin/UhdmAst.cc
mandrys: Thanks! I applied this modification in the newest commit.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3587,8 +3587,12 @@ void UhdmAst::process_assignment_pattern_op()
...
             if (node->type == AST::AST_ASSIGN || node->type == AST::AST_ASSIGN_EQ || node->type == AST::AST_ASSIGN_LE) {
                 // Find at what position in the concat should we place this node
-                auto key = node->children[0]->str;
-                key = key.substr(key.find('.') + 1);
+                std::string key;
+                if (node->children[0]->children[0]->type == static_cast<AST::AstNodeType>(AST::Extended::AST_DOT)) {
+                    key = node->children[0]->children[0]->str;
+                } else {
+                    key = node->children[0]->str;
+                }
```
kamilrakoczy, line 3595: Please add guards to make sure correct nodes exists, e.g.:
```suggestion
                if (!node->children.empty() && !node->children[0]->children.empty() && node->children[0]->children[0]->type == static_cast<AST::AstNodeType>(AST::Extended::AST_DOT)) {
                    key = node->children[0]->children[0]->str;
                } else if (!node->children.empty()) {
                    key = node->children[0]->str;
                } else {
                    log_file_error(node->filename, node->location.first_line,
                               "Couldn't find `key` in assignment pattern.\n");
                }
```

kamilrakoczy (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3587,8 +3587,15 @@ void UhdmAst::process_assignment_pattern_op()
         visit_one_to_many({vpiOperand}, obj_h, [&](AST::AstNode *node) {
             if (node->type == AST::AST_ASSIGN || node->type == AST::AST_ASSIGN_EQ || node->type == AST::AST_ASSIGN_LE) {
                 // Find at what position in the concat should we place this node
```
wsipak, line 3589: Could you please modify this comment or add another one here? Honestly I didn't really help me understand what happens here, and I think it would be beneficial for us to have an explanation of what happens for each of the `vpiOperand`.

mandrys: I added some comments, PTAL to let me know if it makes the operations more understandable.

wsipak: Thanks a lot!

wsipak (approved):

mandrys:
Recent test yosys_systemverilog CI run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4753357780.

PR #490, opened 2023-04-19
Add version to Makefile; use that to automatically tag the release

hzeller:
This is a suggestion to provide a version number to the project, and use that to also tag with the same version in git.

Context:
https://github.com/chipsalliance/yosys-f4pga-plugins/issues/489

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,3 +78,35 @@ jobs:
       env:
         OS: ${{ runner.os }}
         PLUGIN_NAME: ${{ matrix.plugin }}
+
+  release-tagging:
+    name: Version Tagging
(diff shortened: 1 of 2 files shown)


hzeller:
So to tag a new version, just the version number has to be changed in the Makefile and submitted, the rest is automatic.

hzeller:
We're using this succesfully in UHDM and Surelog now so that it is packagable. The CMakeLists.txt contains the version of the project anyway, and so this is then used to derive the current tag.

hzeller:
Any decision on this ?

kgugala (approved):
I'm OK with tagging in Makefile (manually) and then using a CI to create the tag in repository and emit a release

PR #492, opened 2023-04-24
systemverilog-plugin: fix segmentation fault when parameter type is not used in module 

kamilrakoczy:
Fixes: https://github.com/chipsalliance/yosys-f4pga-plugins/issues/491

Parameter types are mostly already propagated by Surelog, we only need them to properly rename module or move typedef definition when they are in modules. This PR skips type parameters when we visit parameters defined outside of module.

This PR also fixes initialization of attributes strings when read command is called multiple times.

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4786916798

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1895,6 +1895,14 @@ void UhdmAst::process_design()
                               shared.top_nodes[node->str] = node;
                           }
                       });
+    visit_one_to_many({vpiParameter, vpiParamAssign}, obj_h, [&](AST::AstNode *node) {
+        if (get_attribute(node, attr_id::is_type_parameter)) {
+            // Don't process type parameters.
+            delete node;
+            return;
+        }
+        add_or_replace_child(current_node, node);
+    });
     visit_one_to_many({vpiTypedef}, obj_h, [&](AST::AstNode *node) {
         if (node)
             move_type_to_new_typedef(current_node, node);
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -1894,7 +1894,6 @@ void UhdmAst::process_design()
                               shared.top_nodes[node->str] = node;
                           }
                       });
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -2042,8 +2043,7 @@ void UhdmAst::process_module()
                         delete node;
                         return;
                     }
-                    if ((node->type == AST::AST_ASSIGN && node->children.size() < 2) ||
-                        (node->type == AST::AST_PARAMETER && get_attribute(node, attr_id::is_type_parameter))) {
```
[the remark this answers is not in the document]
kamilrakoczy: I've moved this condition to separate ``if`` above, but I removed condition about node ``type``. ``attr_id::is_type_parameter`` attribute can only be added to parameters, but it shouldn't matter if it is regular parameter or localparam (we want to delete both).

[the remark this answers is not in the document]
mandrys: Ach, I get it. I missed the change below, thanks!

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1894,7 +1895,6 @@ void UhdmAst::process_design()
                               shared.top_nodes[node->str] = node;
                           }
                       });
-    visit_one_to_many({vpiParameter, vpiParamAssign}, obj_h, [&](AST::AstNode *node) {});
```
mandrys, line 1897: Isn't it needed for processing children of parameters and parameter assignments on file level?

kamilrakoczy: The problem was that it was never added to the design node anyway (so it didn't work), but you are right, I've added support for this instead. New yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4913504824

mandrys: Ok, nice!

mandrys (approved):

PR #494, opened 2023-04-25
XDC: Add missing OBUFT primitive

### LINKED ISSUE #403: IO_LOC_PAIRS not set for OBUFT with T input connected to net
I tried to use 3-state outputs in a project, but the outputs are not driven at all. Setting `T` to 0 drives the outputs as expected.

I created a minimal example on a basys3 board with two switches to reproduce this issue.

Looking through the files revealed that `.IO_LOC_PAIRS` is set to `"NONE"` for the affected `OBUFT`, in the `.eblif` file.
Manually setting `.IO_LOC_PAIRS` on the `OBUFT` makes it work as expected. I verified that the output is floating with a voltmeter.

I attached a archive with the example, a f4pga flow config, my `conda list` output, the `.eblif` files and yosys logfiles of the working and broken state.
[example.tar.gz](https://github.com/chipsalliance/yosys-f4pga-plugins/files/10052571/example.tar.gz)

hannesha:
Fixes #403 

The `OBUFT` primitive was just missing in the xdc plugin.

### DIFF
--- a/xdc-plugin/xdc.cc
+++ b/xdc-plugin/xdc.cc
@@ -56,6 +56,7 @@ const std::unordered_map<std::string, SetPropertyOptions> set_property_options_m
 // corresponding PADs
 const std::unordered_map<std::string, std::vector<std::string>> supported_primitive_parameters = {
   {"OBUF", {"IO_LOC_PAIRS", "IOSTANDARD", "DRIVE", "SLEW", "IN_TERM"}},
+  {"OBUFT", {"IO_LOC_PAIRS", "IOSTANDARD", "DRIVE", "SLEW", "IN_TERM"}},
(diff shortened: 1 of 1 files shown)


mithro:
@kgugala - Who should look at this?

tmichalak:
@hannesha thanks a lot for the fix. Could you enhance the [io_loc_pairs](https://github.com/chipsalliance/yosys-f4pga-plugins/tree/main/xdc-plugin/tests/io_loc_pairs) to cover the OBUFT case? Once the test is in place we shall merge this PR.

hannesha:
@tmichalak Can I use any arbitrary output location or does it have to be a specific pin? Should I add the buffer on multiple levels like the OBUFs?

PR #495, opened 2023-04-26
Make GTEST location configurable.

hzeller:
GTEST_LIBDIR and GTEST_INCDIR are now configurable as parameters to make, which makes it easier to use the gtest library already on the system and not needing the one from third_party/.

Also: invoke yosys from YOSYS_PATH instead of just system `yosys`, as these might be two different things.


hzeller:
I like the solution WIP in #485 by @timkpaine to remove that awkward local building of gtest entirely. Better way to do it.

Tim, might be useful to break out as separate PR.

Will close my PR here.

PR #499, opened 2023-04-27
Improve right-to-left streaming operator support.

mglb:
Fixes streaming operator in black-parrot and opentitan.
Some extra refactoring and fixing as a bonus. Commit-by-commit review is recommended.

UHDM-integration-tests: XXX

yosys-systemverilog: [before passlists update](https://github.com/antmicro/yosys-systemverilog/actions/runs/4812236189), [after passlists update](https://github.com/antmicro/yosys-systemverilog/actions/runs/4813850150)

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3477,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
 
 void UhdmAst::process_stream_op()
 {
-    // Create a for loop that does what a streaming operator would do
-    auto block_node = find_ancestor({AST::AST_BLOCK, AST::AST_ALWAYS, AST::AST_INITIAL});
-    auto process_node = find_ancestor({AST::AST_ALWAYS, AST::AST_INITIAL});
-    auto module_node = find_ancestor({AST::AST_MODULE, AST::AST_FUNCTION, AST::AST_PACKAGE});
-    log_assert(module_node);
-    if (!process_node) {
-        if (module_node->type != AST::AST_FUNCTION) {
-            // Create a @* always block
-            process_node = make_ast_node(AST::AST_ALWAYS);
-            module_node->children.push_back(process_node);
-            block_node = make_ast_node(AST::AST_BLOCK);
-            process_node->children.push_back(block_node);
-        } else {
-            // Create only block
-            block_node = make_ast_node(AST::AST_BLOCK);
-            module_node->children.push_back(block_node);
-        }
-    }
-
-    auto loop_id = shared.next_loop_id();
-    auto loop_counter =
-      make_ast_node(AST::AST_WIRE, {make_ast_node(AST::AST_RANGE, {AST::AstNode::mkconst_int(31, false), AST::AstNode::mkconst_int(0, false)})});
-    loop_counter->is_reg = true;
-    loop_counter->is_signed = true;
-    loop_counter->str = "\\loop" + std::to_string(loop_id) + "::i";
-    module_node->children.insert(module_node->children.end() - 1, loop_counter);
-    auto loop_counter_ident = make_ast_node(AST::AST_IDENTIFIER);
-    loop_counter_ident->str = loop_counter->str;
-
-    auto lhs_node = find_ancestor({AST::AST_ASSIGN, AST::AST_ASSIGN_EQ, AST::AST_ASSIGN_LE})->children[0];
-    // Temp var to allow concatenation
-    AST::AstNode *temp_var = nullptr;
-    AST::AstNode *bits_call = nullptr;
-    if (lhs_node->type == AST::AST_WIRE) {
-        module_node->children.insert(module_node->children.begin(), lhs_node->clone());
-        temp_var = lhs_node->clone(); // if we already have wire as lhs, we want to create the same wire for temp_var
-        lhs_node->delete_children();
-        lhs_node->type = AST::AST_IDENTIFIER;
-        bits_call = make_ast_node(AST::AST_FCALL, {lhs_node->clone()});
-        bits_call->str = "\\$bits";
-    } else {
-        // otherwise, we need to calculate size using bits fcall
-        bits_call = make_ast_node(AST::AST_FCALL, {lhs_node->clone()});
-        bits_call->str = "\\$bits";
-        temp_var =
-          make_ast_node(AST::AST_WIRE, {make_ast_node(AST::AST_RANGE, {make_ast_node(AST::AST_SUB, {bits_call, AST::AstNode::mkconst_int(1, false)}),
-                                                                       AST::AstNode::mkconst_int(0, false)})});
-    }
-
-    temp_var->str = "\\loop" + std::to_string(loop_id) + "::temp";
-    module_node->children.insert(module_node->children.end() - 1, temp_var);
-    auto temp_var_ident = make_ast_node(AST::AST_IDENTIFIER);
-    temp_var_ident->str = temp_var->str;
-    auto temp_assign = make_ast_node(AST::AST_ASSIGN_EQ, {temp_var_ident});
-    block_node->children.push_back(temp_assign);
-
-    // Assignment in the loop's block
-    auto assign_node = make_ast_node(AST::AST_ASSIGN_EQ, {lhs_node->clone(), temp_var_ident->clone()});
-    AST::AstNode *slice_size = nullptr; // First argument in streaming op
-    visit_one_to_many({vpiOperand}, obj_h, [&](AST::AstNode *node) {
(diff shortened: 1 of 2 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -3477,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
...
+        (make_ident(src_wire->str)),
+        (stream_concat_arg),
+      });
+
+    AST::AstNode *const loop_counter = //
+      (make_node(is_proc_ctx ? AST::AST_WIRE : AST::AST_GENVAR).str(make_id_str("counter")).is_reg(true))({
+        (make_range(31, 0, true)),
+      });
+
+    AST::AstNode *const for_loop = //
```
kamilrakoczy, line 3574: This will be very hard to understood later. I think this needs to be spitted at least to ``loop_header``, ``loop_body`` and maybe ``block`` and ``assign``.

It also requires some comment what it is actually doing.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3408,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
...
+      AST::AST_BLOCK,
+      AST::AST_INITIAL,
+      AST::AST_ALWAYS,
+      AST::AST_FUNCTION,
+    });
+    log_assert(stmt_list_node != nullptr);
+
+    // Detect whether we're in a procedural context. If yes, `for` loop will be generated, and `generate for` otherwise.
+    const AST::AstNode *const proc_ctx = find_ancestor({AST::AST_ALWAYS, AST::AST_INITIAL, AST::AST_FUNCTION});
+    const bool is_proc_ctx = (proc_ctx != nullptr);
```
kamilrakoczy, line 3493: Maybe this isn't nice, but we don't need to traverse tree 2 times.
```suggestion
    const bool is_proc_ctx = (stmt_list_node->type != AST::AST_ALWAYS || stmt_list_node->type != AST::AST_INITIAL || stmt_list_node->type != AST::AST_FUNCTION);
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3408,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
...
+    // Closest ancestor where new statements can be inserted.
+    AST::AstNode *stmt_list_node = find_ancestor({
+      AST::AST_MODULE,
+      AST::AST_PACKAGE,
+      AST::AST_BLOCK,
+      AST::AST_INITIAL,
+      AST::AST_ALWAYS,
+      AST::AST_FUNCTION,
+    });
+    log_assert(stmt_list_node != nullptr);
```
kamilrakoczy, line 3489: Please change this to ``log_file_error``, that logs also filename and line number as we might sometimes hit this assert.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3408,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
...
-    auto loop_node = make_ast_node(AST::AST_FOR);
-    loop_node->str = "$loop" + std::to_string(loop_id);
-    loop_node->children.push_back(init_stmt);
-    loop_node->children.push_back(cond_stmt);
-    loop_node->children.push_back(inc_stmt);
-    loop_node->children.push_back(make_ast_node(AST::AST_BLOCK, {assign_node}));
-    loop_node->children[3]->str = "\\stream_op_block" + std::to_string(loop_id);
-
-    block_node->children.push_back(make_ast_node(AST::AST_BLOCK, {loop_node}));
+        stream_concat_arg = operands.back();
```
kamilrakoczy, line 3529: We are asserting size, so it probably makes sens? I don't have hard opinion so I'm fine with both.
```suggestion
        stream_concat_arg = operands.at(1);
```

kamilrakoczy (approved):
LGTM, only minor comments and suggestions.

systemverilog-plugin/UhdmAst.cc
mglb: No, it must be `back()`. The first operand is optional, so `operands` can have either one or two entries. I guess this huge removed block in the diff obscured things, but there is an if/else just above this statement that handles either 2 or 1 operand.

systemverilog-plugin/UhdmAst.cc
mglb: Can't do that. `stmt_list_node` could be an `AST_BLOCK`. We have to check whether this AST_BLOCK is inside AST_ALWAYS/INITIAL/FUNCTION or not.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3477,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
...
+          (make_node(AST::AST_SUB))({
+            (make_ident(stream_concat_width_lp->str)),
+            (make_const(1)),
+          }),
+          (make_const(0)),
+        }),
+      });
+
+    AST::AstNode *const assign_stream_concat_to_src_wire = //
+      (make_node(is_proc_ctx ? AST::AST_ASSIGN_EQ : AST::AST_ASSIGN))({
```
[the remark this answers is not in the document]
mglb: I, personally, would find such code harder to read.

When reading the code now the reader have to keep only one boolean in mind (`in_proc_ctx`).
They can also just visually scan the code (which intentionally mimics actual tree dump) and quickly see node types.
Even without knowing what `in_proc_ctx` is, it is visible in-place that this is either node of type X or Y.

With your suggestion the reader would need to keep in mind multiple non-boolean state variables - `for_or_genfor`, `assign_or_assign_eq`, `wire_or_genvar`.
It would be harder to catch node types during quick visual scan of the code, because one would need to look either for a variable or a type.

I would consider it as a better option in one case - when the in-place expression would cause the line to wrap. But since we have pretty big column limit we can make use of it.

systemverilog-plugin/UhdmAst.cc
mglb: I agree with "will be hard to understand". However, to keep it linear and structurally as close to the actual tree dump as possible, what do you think about marking what is what with comments? Like: https://gist.github.com/mglb/54be7f2e64fea7c5703ae98d24c77dbe

As for comment, do you mean something like "this implements the streaming operator as a for loop which assigns expression slices to an internal wire/variable in expected order, then emits reference to the wire/variable in place of actual streaming operator.", or description what the tree describing code actually generates?
I hope not the latter, as the sole point of the builder class is to make this code as self-describing as possible :D

systemverilog-plugin/UhdmAst.cc
mglb: I see the point, but to not mix assertions with errors I've added yet another thing to this PR: `uhdmast_assert` (see commits).
It prints errors like:
```
/yosys-systemverilog/yosys-f4pga-plugins/systemverilog-plugin/UhdmAst.cc:3536: error: Assertion failed: stmt_list_node != nullptr
/yosys-systemverilog/yosys-f4pga-plugins/systemverilog-plugin/UhdmAst.cc:3536: note: In function: void systemverilog_plugin::UhdmAst::process_stream_op()
/yosys-systemverilog/UHDM-integration-tests/tests/StreamOpInitialAlways/dut.sv:2018: note: When processing object of type 'operation'.
```

Ofc I've used it here and in the rest of the method.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -3408,120 +3477,171 @@ void UhdmAst::process_operation(const UHDM::BaseClass *object)
...
-            slice_size = node;
+    // TODO (mglb): Only concat expression's size factors are supported as a slice size. Add support for other slice sizes as well.
+    AST::AstNode *slice_size_arg = nullptr;
+    AST::AstNode *stream_concat_arg = nullptr;
+    {
+        std::vector<AST::AstNode *> operands;
+        // Expected operands: [slice_size] stream_concatenation
+        visit_one_to_many({vpiOperand}, obj_h, [&](AST::AstNode *node) {
+            log_assert(node != nullptr);
+            log_assert(operands.size() < 2);
```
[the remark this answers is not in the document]
mglb: I've removed the one below to detect errors early.

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: I meant something that you proposed as gist, it is enough for me.

systemverilog-plugin/UhdmAst.cc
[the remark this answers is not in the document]
kamilrakoczy: I don't have strong opinion and I see your point, so lets not change this.

PR #506, opened 2023-05-08
sv-plugin: Strip the node name using parent name instead of "."

mandrys:
Considering the fact that the name can consist of escaped characters, including dot, we need to change the way of stripping the name of a node - instead of a dot, we now use the ancestor name. 

Added test: https://github.com/chipsalliance/UHDM-integration-tests/pull/730 (It fails formal verification because of sv2v being unable to handle escaped characters, but is useful as a parsing test.)

Test yosys-systemverilog CI run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4916684728.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -174,6 +174,43 @@ static void sanitize_symbol_name(std::string &name)
     }
 }
 
+static std::string get_parent_name(vpiHandle parent_h)
+{
+    std::string parent_name;
+    if (auto p = vpi_get_str(vpiFullName, parent_h)) {
+        parent_name = p;
+    } else if (auto p = vpi_get_str(vpiName, parent_h)) {
+        parent_name = p;
+    } else if (auto p = vpi_get_str(vpiDefName, parent_h)) {
+        parent_name = p;
+    }
+    return parent_name;
+}
+
+// Warning: Takes ownership of `parent_h` and releases it.
+static void find_ancestor_name(vpiHandle parent_h, std::string &name, std::string &parent_name)
+{
+
+    while (!parent_name.empty()) {
+        parent_name = parent_name + ".";
+        if ((name.rfind(parent_name) != std::string::npos)) {
+            name = name.substr(name.rfind(parent_name) + parent_name.size());
+            break;
+        } else {
+            auto old_parent_h = parent_h;
+            parent_h = vpi_handle(vpiParent, parent_h);
+            vpi_release_handle(old_parent_h);
+
+            if (parent_h) {
+                parent_name = get_parent_name(parent_h);
+            } else {
+                parent_name.clear();
+            }
+        }
+    }
+    vpi_release_handle(parent_h);
+}
+
 static std::string get_name(vpiHandle obj_h, bool prefer_full_name = false)
 {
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -183,8 +183,75 @@ static std::string get_name(vpiHandle obj_h, bool prefer_full_name = false)
...
+            parent_h = vpi_handle(vpiParent, parent_h);
+            if (parent_h) {
+                if (auto p = vpi_get_str(vpiFullName, parent_h)) {
+                    parent_name = p;
+                } else if (auto p = vpi_get_str(vpiName, parent_h)) {
+                    parent_name = p;
+                } else if (auto p = vpi_get_str(vpiDefName, parent_h)) {
+                    parent_name = p;
+                }
+                while (!parent_name.empty()) {
```
kamilrakoczy, line 213: I think this requires some refactoring as there is a lot of code repetition. Please extract recurrent patterns to separate functions.

kamilrakoczy (changes requested):

wsipak:
We've had an offline discussion about the `vpi_handle`s and the other API of UHDM that allows dealing with objects, that we could consider here. I think using the `UHDM::BaseClass` -based stuff could also help us keep the code cleaner.

systemverilog-plugin/UhdmAst.cc

mandrys:
New _yosys-systemverilog_ run: https://github.com/antmicro/yosys-systemverilog/actions/runs/5134240233.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -186,9 +219,30 @@ static std::string get_name(vpiHandle obj_h, bool prefer_full_name = false)
...
+        parent_name = get_parent_name(parent_h);
+
+        if (parent_name.empty()) {
+            // Nodes of certain types, like param_assign, don't have
+            // a name, so we need to look further for the ancestor.
+            parent_h = vpi_handle(vpiParent, parent_h);
+
+            if (parent_h) {
+                parent_name = get_parent_name(parent_h);
+                find_ancestor_name(parent_h, name, parent_name);
```
mglb, line 240:
```suggestion
```
I think `find_ancestor_name` call below will handle it. Is this correct?

systemverilog-plugin/UhdmAst.cc
```diff
@@ -186,9 +219,30 @@ static std::string get_name(vpiHandle obj_h, bool prefer_full_name = false)
...
+            }
+        }
+        find_ancestor_name(parent_h, name, parent_name);
     }
+    vpi_release_handle(parent_h);
```
mglb, line 245:
```suggestion
```
Handle released in find_ancestor_name above.

mglb (changes requested):
Few changes; please re-run yosys-systemverilog with it. If it passes it will mean I didn't make any mistake.

mandrys:
Latest _yosys-systemverilog_ run passed, only formal verification failed for some unexpected passes: https://github.com/antmicro/yosys-systemverilog/actions/runs/5134720743/jobs/9239550297.
When the PR is approved, I'll squash the commits with fixes and merge it.

mglb (approved):

PR #507, opened 2023-05-10
systemverilog-plugin: fix packed arrays of enums

kamilrakoczy:
This PR fixes size and access to packed array of enum.

UHDM-integration-test: https://github.com/chipsalliance/UHDM-integration-tests/pull/732
yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/4935744732

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -496,6 +496,8 @@ static void resolve_wiretype(AST::AstNode *wire_node)
             unpacked_ranges.push_back(r->clone());
         }
     }
+    delete_attribute(wire_node, attr_id::packed_ranges);
+    delete_attribute(wire_node, attr_id::unpacked_ranges);
     AST::AstNode *wiretype_ast = nullptr;
     log_assert(AST_INTERNAL::current_scope.count(wiretype_node->str));
     wiretype_ast = AST_INTERNAL::current_scope[wiretype_node->str];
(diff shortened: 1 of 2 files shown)


wsipak (approved):
We could make use of some comments in this function if you have any idea for them. For example, I think it would be beneficial to explain why we created `[un]packed_ranges_wiretype` vectors instead of just pushing to `[un]packed_ranges`. I can see that this is somehow related to the order of elements in the vectors.
Anyways, LGTM!

systemverilog-plugin/UhdmAst.cc
```diff
@@ -1758,17 +1747,15 @@ void UhdmAst::process_packed_array_typespec()
...
-                    if (c->type == AST::AST_RANGE && c->str.empty())
-                        packed_ranges.push_back(c);
-                    else
-                        delete c;
-                }
-                node->children[0]->children.clear();
+            if (!node->str.empty()) {
+                auto wiretype_node = make_ast_node(AST::AST_WIRETYPE);
+                wiretype_node->str = node->str;
+                current_node->children.push_back(wiretype_node);
```
mglb, line 1753:
```suggestion
                AST::AstNode *const wiretype_node = make_named_node(AST::AST_WIRETYPE);
                wiretype_node->str = node->str;
                current_node->children.push_back(wiretype_node);
```
Let's use `make_named_node` instead of `make_ast_node` everywhere, with the intention of removing `make_ast_node` someday. They do the same thing, but the former returns `AstNodeBuilder`, which makes it more flexible.

mglb (approved):
Just one request. Please fix before merging.

mandrys (approved):

kamilrakoczy:
Actually this change makes opentitan to timeout in CI, marking as draft until resolved.

kamilrakoczy:
New yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/5176142077
now all tests passes.

mglb (approved):

PR #511, opened 2023-05-12
Process vpiArrayTypespec the same way as vpiPackedArrayTypespec

wsipak:
Previously the was no case for `vpiArrayTypespec` in `process_typespec_member`.

CI run here: https://github.com/antmicro/yosys-systemverilog/actions/runs/4958803078


kamilrakoczy:
Information that we were interested in is now reported as ``vpiPackedArrayTypespec`` (https://github.com/chipsalliance/Surelog/issues/3642#issuecomment-1548959540), so lets close this PR for now and reopen when we will need to handle ``vpiArrayTypespec``.

PR #512, opened 2023-05-15
systemverilog-plugin: update remove module message

kamilrakoczy:
This PR updates message about removing unused module.

Previous message could be misleading that whole module was removed, but we are removing only unelaborated (and unused) modules.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -2012,7 +2012,7 @@ void UhdmAst::process_design()
                 current_node->children.push_back(pair.second);
             }
         } else {
-            log_warning("Removing unused module: %s from the design.\n", pair.second->str.c_str());
+            log_warning("Removing unelaborated module: %s from the design.\n", pair.second->str.c_str());
             // TODO: This should be properly erased from the module, but it seems that it's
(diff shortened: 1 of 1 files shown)


mglb (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -2012,7 +2012,7 @@ void UhdmAst::process_design()
                 current_node->children.push_back(pair.second);
             }
         } else {
-            log_warning("Removing unused module: %s from the design.\n", pair.second->str.c_str());
+            log_warning("Removing unparameterized module: %s from the design.\n", pair.second->str.c_str());
```
mandrys, line 2015: What does it actually mean that the module is unparameterized?

kamilrakoczy: This is ``unelaborated`` model from Surelog, so one that doesn't have parameters replaced, genscopes unrolled etc.

I actually changed message to ``unelaborated``.

mandrys: Ok, now it's clear, thanks!

mandrys (approved):

mglb (approved):

PR #520, opened 2023-05-25
systemverilog-plugin: allow multirange access in dot access

wsipak:
See https://github.com/chipsalliance/yosys-f4pga-plugins/issues/517

In order to allow multiranges in dot notation, it's useful to first fix the usage of `multirange_dimensions` in the plugin.
See https://github.com/YosysHQ/yosys/blob/master/frontends/ast/ast.h#L205
Currently, the plugin doesn't put proper data into the vector. For objects of type `AST_STRUCT_ITEM` it puts there only one number representing the size of the first dimension.
With these changes, the function `add_multirange_attribute` will always push even number of elements to `multirange_dimensions`, which is the way it should work. It also handles `AST_STRUCT_ITEMS` so that information about all the dimensions is included, not only the first one.

The function `expand_dot` will now allow multiple ranges, and will iterate through them to calculate a proper range to access the selected multi-range wire.

See CI run here:
https://github.com/antmicro/yosys-systemverilog/actions/runs/5130614569

A new test here:
https://github.com/chipsalliance/UHDM-integration-tests/pull/737
CI run with the new test:
https://github.com/antmicro/yosys-systemverilog/actions/runs/5131504505

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -295,6 +295,11 @@ static void add_multirange_wire(AST::AstNode *node, std::vector<AST::AstNode *> 
 
 static size_t add_multirange_attribute(AST::AstNode *wire_node, const std::vector<AST::AstNode *> ranges)
 {
+    // node->multirange_dimensions stores dimensions' offsets and widths.
+    // It shall have even number of elements.
+    // For a range of [A:B] it should be appended with {min(A,B)} and {max(A,B)-min(A,B)+1}
+    // For a range of [A] it should be appended with {0} and {A}
+
     size_t size = 1;
     for (size_t i = 0; i < ranges.size(); i++) {
         log_assert(AST_INTERNAL::current_ast_mod);
@@ -320,17 +325,15 @@ static size_t add_multirange_attribute(AST::AstNode *wire_node, const std::vecto
         }
         log_assert(ranges[i]->children[0]->type == AST::AST_CONSTANT);
         log_assert(ranges[i]->children[1]->type == AST::AST_CONSTANT);
-        if (wire_node->type != AST::AST_STRUCT_ITEM) {
-            wire_node->multirange_dimensions.push_back(min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer));
-            wire_node->multirange_swapped.push_back(ranges[i]->range_swapped);
-        }
         auto elem_size = max(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer) -
                          min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer) + 1;
-        if (wire_node->type != AST::AST_STRUCT_ITEM || (wire_node->type == AST::AST_STRUCT_ITEM && i == 0)) {
-            wire_node->multirange_dimensions.push_back(elem_size);
-        }
+        wire_node->multirange_dimensions.push_back(min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer));
+        wire_node->multirange_dimensions.push_back(elem_size);
+        wire_node->multirange_swapped.push_back(ranges[i]->range_swapped);
         size *= elem_size;
     }
+    log_assert(wire_node->multirange_dimensions.size() % 2 == 0);
+
     return size;
 }
 
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -325,9 +325,12 @@ static size_t add_multirange_attribute(AST::AstNode *wire_node, const std::vecto
         }
         log_assert(ranges[i]->children[0]->type == AST::AST_CONSTANT);
         log_assert(ranges[i]->children[1]->type == AST::AST_CONSTANT);
-        auto elem_size = max(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer) -
-                         min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer) + 1;
-        wire_node->multirange_dimensions.push_back(min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer));
+
+        const auto low = min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer);
+        const auto high = max(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer);
+        const auto elem_size = high - low + 1;
+
+        wire_node->multirange_dimensions.push_back(low);
         wire_node->multirange_dimensions.push_back(elem_size);
         wire_node->multirange_swapped.push_back(ranges[i]->range_swapped);
         size *= elem_size;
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -780,7 +779,7 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
         // Currently support only special access to 2 dimensional packed element
         // when selecting single range
         log_assert(current_struct_elem->multirange_dimensions.size() % 2 == 0);
-        if (struct_range && (current_struct_elem->multirange_dimensions.size() / 2) == 2) {
+        if (struct_ranges.size() && (current_struct_elem->multirange_dimensions.size() / 2) == 2) {
```
mglb, line 782:
```suggestion
        if (!struct_ranges.empty() && (current_struct_elem->multirange_dimensions.size() / 2) == 2) {
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -806,7 +805,19 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
         std::swap(right, sub_dot->children[1]);
         delete sub_dot;
     }
-    if (struct_range) {
+
+    for (size_t i = 0; i < struct_ranges.size(); i++) {
+        auto *struct_range = struct_ranges[i];
```
mglb, line 810:
```suggestion
        const auto * const struct_range = struct_ranges[i];
```
(quick look says it's possible, but please verify)

systemverilog-plugin/UhdmAst.cc
```diff
@@ -806,7 +805,19 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
...
+
+        int range_width = 0;
+        if (!current_struct_elem->multirange_dimensions.size()) {
```
mglb, line 813:
```suggestion
        if (current_struct_elem->multirange_dimensions.empty()) {
```

systemverilog-plugin/UhdmAst.cc
```diff
@@ -806,7 +805,19 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
...
+            range_width = 1;
+        } else if (current_struct_elem->multirange_dimensions.size() > (i * 2 + 1)) {
+            range_width = current_struct_elem->multirange_dimensions[i * 2 + 1];
+            // TODO consider offsets for ranges that do not start from 0. Right now every offset is assumed 0.
```
mglb, line 817:
```c++
const auto range_offset = current_struct_elem->multirange_dimensions[i * 2];
if (range_offset != 0) {
    log_file_error(...)
}
```

mglb (commented):
Stylistic comments, functionally looks OK.
Please wait for @kamilrakoczy review, I think he knows `expand_dot` better than me.

mglb (changes requested):
requesting changes

systemverilog-plugin/UhdmAst.cc
```diff
@@ -320,17 +325,15 @@ static size_t add_multirange_attribute(AST::AstNode *wire_node, const std::vecto
...
-            wire_node->multirange_swapped.push_back(ranges[i]->range_swapped);
-        }
         auto elem_size = max(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer) -
                          min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer) + 1;
-        if (wire_node->type != AST::AST_STRUCT_ITEM || (wire_node->type == AST::AST_STRUCT_ITEM && i == 0)) {
-            wire_node->multirange_dimensions.push_back(elem_size);
-        }
+        wire_node->multirange_dimensions.push_back(min(ranges[i]->children[0]->integer, ranges[i]->children[1]->integer));
+        wire_node->multirange_dimensions.push_back(elem_size);
+        wire_node->multirange_swapped.push_back(ranges[i]->range_swapped);
```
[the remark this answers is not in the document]
wsipak: This is a good idea, done!

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc
wsipak: LGTM, I've applied this.

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc

systemverilog-plugin/UhdmAst.cc
```diff
@@ -295,6 +295,11 @@ static void add_multirange_wire(AST::AstNode *node, std::vector<AST::AstNode *>
 
 static size_t add_multirange_attribute(AST::AstNode *wire_node, const std::vector<AST::AstNode *> ranges)
 {
+    // node->multirange_dimensions stores dimensions' offsets and widths.
+    // It shall have even number of elements.
+    // For a range of [A:B] it should be appended with {min(A,B)} and {max(A,B)-min(A,B)+1}
+    // For a range of [A] it should be appended with {0} and {A}
+
```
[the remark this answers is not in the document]
wsipak: Right now it's not safe to assert this, I think we could warn and return if `multirange_dimensions` already exist, but I need to delve into this first.

kamilrakoczy (approved):
Generally LGTM, only some minor comments.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -806,7 +808,23 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
...
+        } else if (current_struct_elem->multirange_dimensions.size() > (i * 2 + 1)) {
+            range_width = current_struct_elem->multirange_dimensions[i * 2 + 1];
+            const auto range_offset = current_struct_elem->multirange_dimensions[i * 2];
+            if (range_offset != 0) {
+                log_file_error(struct_range->filename, struct_range->location.first_line,
+                               "Accessing ranges that do not start from 0 is not supported.");
+            }
+        } else {
+            break;
+        }
```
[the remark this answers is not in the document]
wsipak: I'm sorry, I don't think I understand what you mean.
The changes I'm applying here allow selecting more than one position, as in `foo[x][1:0]` the result will be 2 positions (and this works for the test case). It's illegal to select `foo[x:y][a:b]`, I think.

Can you explain what case I should additionally handle here?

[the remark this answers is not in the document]
kamilrakoczy: I've used following test:
```systemverilog
typedef struct packed {logic [3:0][7:0] k;} my_struct_packed_t;

module single_range_dot_access (
    input  my_struct_packed_t x1,
    output my_struct_packed_t y1,
    input  my_struct_packed_t x2,
    output my_struct_packed_t y2
);
  assign y1.k[0][0:1] = x1.k[1][0:1];
endmodule : single_range_dot_access
```
and as a result I have:
```systemverilog
assign y1[7:0] = x1[-16:9];
```
this should be valid systemverilog (I've used Surelog version ``c9dbd09e64a596297f7715987c6ebdd2f8bd4753`` in my tests).

systemverilog-plugin/UhdmAst.cc
```diff
@@ -806,7 +808,23 @@ static AST::AstNode *expand_dot(const AST::AstNode *current_struct, const AST::A
...
+            range_width = 1;
+        } else if (current_struct_elem->multirange_dimensions.size() > (i * 2 + 1)) {
+            range_width = current_struct_elem->multirange_dimensions[i * 2 + 1];
+            const auto range_offset = current_struct_elem->multirange_dimensions[i * 2];
+            if (range_offset != 0) {
+                log_file_error(struct_range->filename, struct_range->location.first_line,
+                               "Accessing ranges that do not start from 0 is not supported.");
+            }
+        } else {
+            break;
```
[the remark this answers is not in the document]
wsipak: Alright, this is done.

PR #522, opened 2023-05-30
antmicro/yosys-systemverilog#1743: Compensate for chipsalliance/Surelog#3670

hs-apotell:
antmicro/yosys-systemverilog#1743: Compensate for chipsalliance/Surelog#3670

UHDM model hierarchy changed to enforce vpiParent as weak reference.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -3985,9 +3985,6 @@ void UhdmAst::process_bit_select()
 void UhdmAst::process_part_select()
 {
     current_node = make_ast_node(AST::AST_IDENTIFIER);
-    vpiHandle parent_h = vpi_handle(vpiParent, obj_h);
-    current_node->str = get_name(parent_h);
-    vpi_release_handle(parent_h);
     auto range_node = new AST::AstNode(AST::AST_RANGE);
     range_node->filename = current_node->filename;
     range_node->location = current_node->location;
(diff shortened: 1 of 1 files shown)


mglb:
I did run the PR + https://github.com/chipsalliance/Surelog/pull/3670 in yosys-systemverilog CI, many tests fail: https://github.com/antmicro/yosys-systemverilog/actions/runs/5121636527
I'll look into the Surelog/UHDM change soon to see what should be done in the plugin.

mglb:
@hs-apotell
Turns out the plugin doesn't work with Surelog's master, which is responsible for at least some issues from the yosys-systemverilog CI run linked above.
It would be best to wait until we catch up with the master branch - this work is already in progress.

mglb:
@hs-apotell FYI: you can fork [yosys-systemverilog](https://github.com/antmicro/yosys-systemverilog), change submodule revisions/remotes to point to your changes, commit+push, and create a Draft PR. The CI will test it using code from revisions/forks configured on your branch.

Instructions how to change submodule can be found in the readme: https://github.com/antmicro/yosys-systemverilog/pull/1772

mglb:
Every related issue reported against Surelog/UHDM has been fixed AFAIK.
With https://github.com/chipsalliance/yosys-f4pga-plugins/pull/528 merged, only one test in yosys-systemverilog fails with Surelog master.

pgielda:
Custom runners are "private" to an org. In principle the only way for them to run for external users is to open a PR to that org (so in this case to https://github.com/antmicro/yosys-systemverilog). You cannot use them "privately" (unless you set up exactly the same "private" infrastructure)

PR #529, opened 2023-06-06
yosys-systemverilog: fix multirange with dot usage

kamilrakoczy:
Ref: https://github.com/chipsalliance/yosys-f4pga-plugins/issues/524

This PR adds support for multirange in ``dot`` context (before first ``dot``).

UHDM-integration-test: https://github.com/chipsalliance/UHDM-integration-tests/pull/740
yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/5186512253/jobs/9347862380

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -962,28 +962,46 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
     }
     log_assert(struct_node);
     auto expanded = expand_dot(struct_node, dot);
-    if (node->children[0]->type == AST::AST_RANGE) {
-        int struct_size_int = get_max_offset_struct(struct_node) + 1;
-        log_assert(!wire_node->multirange_dimensions.empty());
-        int range = wire_node->multirange_dimensions.back() - 1;
-        if (!wire_node->attributes[UhdmAst::unpacked_ranges()]->children.empty() &&
-            wire_node->attributes[UhdmAst::unpacked_ranges()]->children.back()->range_left == range) {
-            expanded->children[1] = new AST::AstNode(
-              AST::AST_ADD, expanded->children[1],
-              new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32),
-                               new AST::AstNode(AST::AST_SUB, AST::AstNode::mkconst_int(range, true, 32), node->children[0]->children[0]->clone())));
-            expanded->children[0] = new AST::AstNode(
-              AST::AST_ADD, expanded->children[0],
-              new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32),
-                               new AST::AstNode(AST::AST_SUB, AST::AstNode::mkconst_int(range, true, 32), node->children[0]->children[0]->clone())));
-        } else {
-            expanded->children[1] = new AST::AstNode(
-              AST::AST_ADD, expanded->children[1],
-              new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), node->children[0]->children[0]->clone()));
-            expanded->children[0] = new AST::AstNode(
-              AST::AST_ADD, expanded->children[0],
-              new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), node->children[0]->children[0]->clone()));
+    std::vector<AST::AstNode *> struct_ranges;
+    for (auto c : node->children) {
+        if (c->type == AST::AST_RANGE) {
+            struct_ranges.push_back(c);
         }
+    }
+    const auto wire_node_unpacked_ranges_size = wire_node->attributes[UhdmAst::unpacked_ranges()]->children.size();
+    // 'dot' notation can only be used on specific struct element
+    // '1' packed range is from wiretype and it is already expanded in 'expand_dot'
+    // TODO(krak): wire ranges are sometimes under wiretype node (e.g. in case of typedef)
+    // but wiretype ranges contains also struct range that is already expanded in 'expand_dot'
+    // we need to find a way to calculate size of wire ranges without struct range here to enable this assert
+    // const auto wire_node_packed_ranges_size = wire_node->attributes[UhdmAst::packed_ranges()]->children.size();
+    // const auto wire_node_ranges_size = wire_node_packed_ranges_size + wire_node_unpacked_ranges_size;
+    // log_assert(struct_ranges.size() == (wire_node_ranges_size - 1));
+
+    // Get size of single structure
+    int struct_size_int = get_max_offset_struct(struct_node) + 1;
+    // Get reverse iterator of wire dimensions
+    auto dimensions_it = wire_node->multirange_dimensions.rbegin();
+    unsigned long range_id = 0;
+    for (auto it = struct_ranges.rbegin(); it != struct_ranges.rend(); it++) {
+        // in 'dot' context, we need to select specific struct
+        // element, so assert that there is only 1 child (range with single const) in struct
+        log_assert((*it)->children.size() == 1);
+        bool is_unpacked_range = range_id < wire_node_unpacked_ranges_size;
+        // if unpacked range, select from back;
+        auto elem = is_unpacked_range
+                      ? new AST::AstNode(AST::AST_SUB, AST::AstNode::mkconst_int(*dimensions_it - 1, true, 32), (*it)->children[0]->clone())
+                      : (*it)->children[0]->clone();
+        // calculate which struct we selected
+        auto move_offset = new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), elem);
+        // move our expanded dot to currently selected struct
+        expanded->children[0] = new AST::AstNode(AST::AST_ADD, move_offset->clone(), expanded->children[0]);
+        expanded->children[1] = new AST::AstNode(AST::AST_ADD, move_offset, expanded->children[1]);
+        // next range, multiply struct_size_int by whole dimension size
+        struct_size_int *= *dimensions_it;
+        // move iterator to next dimension width
+        dimensions_it += 2;
+        range_id++;
     }
     return expanded;
 }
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,47 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
+                      : (*it)->children[0]->clone();
+        // calculate which struct we selected
+        auto move_offset = new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), elem);
+        // move our expanded dot to currently selected struct
+        expanded->children[0] = new AST::AstNode(AST::AST_ADD, move_offset->clone(), expanded->children[0]);
+        expanded->children[1] = new AST::AstNode(AST::AST_ADD, move_offset, expanded->children[1]);
+        // next range, multiply struct_size_int by whole dimension size
+        struct_size_int *= *dimensions_it;
+        // move iterator to next dimension width
+        dimensions_it += 2;
```
wsipak, line 1003: Shouldn't we check whether `dimensions_it` exceeds the length of `multirage_dimensions`?
We could have something to make sure that we don't violate the size of the vector.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,47 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
-              new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), node->children[0]->children[0]->clone()));
+    std::vector<AST::AstNode *> struct_ranges;
+    for (auto c : node->children) {
+        if (c->type == AST::AST_RANGE) {
+            struct_ranges.push_back(c);
         }
     }
+    const auto wire_node_unpacked_ranges_size = wire_node->attributes[UhdmAst::unpacked_ranges()]->children.size();
+    // 'dot' notation can only be used on specific struct element
+    // '1' packed range is from wiretype and it is already expanded in 'expand_dot'
```
mglb, line 973: What `'1'` means? Is it the first (leftmost) packed range?

mglb (commented):
After the first read it looks OK, but I would like to look at it one more time in a few hours.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,47 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
+        log_assert((*it)->children.size() == 1);
+        bool is_unpacked_range = range_id < wire_node_unpacked_ranges_size;
+        // if unpacked range, select from back;
+        auto elem = is_unpacked_range
+                      ? new AST::AstNode(AST::AST_SUB, AST::AstNode::mkconst_int(*dimensions_it - 1, true, 32), (*it)->children[0]->clone())
+                      : (*it)->children[0]->clone();
+        // calculate which struct we selected
+        auto move_offset = new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), elem);
+        // move our expanded dot to currently selected struct
+        expanded->children[0] = new AST::AstNode(AST::AST_ADD, move_offset->clone(), expanded->children[0]);
```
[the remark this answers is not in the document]
kamilrakoczy: Yes, ``expanded`` is result from ``expand_dot`` that always return ``AST_RANGE`` with 2 children:
https://github.com/chipsalliance/yosys-f4pga-plugins/blob/8edc027cdc484f86c5ac8698c696c74d04e6f92e/systemverilog-plugin/UhdmAst.cc#L955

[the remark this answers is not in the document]
wsipak: Makes sense, thanks!

systemverilog-plugin/UhdmAst.cc
kamilrakoczy: I tried to explain here why there is ``wire_node_ranges_size - 1`` in assert (it was due to packed range that comes from wiretype), but ultimately I've disabled it as it is more complicated in some cases. I think I will remove this comment for now and left only TODO, as it is not valid for all cases.

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,47 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
-            expanded->children[0] = new AST::AstNode(
-              AST::AST_ADD, expanded->children[0],
-              new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), node->children[0]->children[0]->clone()));
+    std::vector<AST::AstNode *> struct_ranges;
+    for (auto c : node->children) {
+        if (c->type == AST::AST_RANGE) {
+            struct_ranges.push_back(c);
         }
     }
+    const auto wire_node_unpacked_ranges_size = wire_node->attributes[UhdmAst::unpacked_ranges()]->children.size();
```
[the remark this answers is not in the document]
kamilrakoczy: At this point, ``wire_node`` should have it, I've added assert to make sure it is always present.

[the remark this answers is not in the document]
wsipak: Thanks!

wsipak (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,49 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
+    // log_assert(struct_ranges.size() == (wire_node_ranges_size - 1));
+
+    // Get size of single structure
+    int struct_size_int = get_max_offset_struct(struct_node) + 1;
+    // Get reverse iterator of wire dimensions
+    auto dimensions_it = wire_node->multirange_dimensions.rbegin();
+    unsigned long range_id = 0;
+    for (auto it = struct_ranges.rbegin(); it != struct_ranges.rend(); it++) {
+        // in 'dot' context, we need to select specific struct
+        // element, so assert that there is only 1 child (range with single const) in struct
```
mandrys, line 990: Here we assert that an item in `struct_ranges` has only 1 child, but the comment says that we expect only 1 child in a struct - is that right or I'm messing sth up? Generally this comment is not clear for me - could you explain, why this assert is important for the fact that we select a specific struct in 'dot' context?

kamilrakoczy: Lets consider such assign:
```systemverilog
assign group_status[0][0].a = 1'b1;
```
At this point, ``.a`` part was already expanded by ``expand_dot`` function, and here we are reverse-iterating ranges of ``group_status``.

This assert checks, that each range have only 1 child, so it is range with single const, not part-select (e.g. ``[1:0]``) as latter is not valid systemverilog code.

I slightly changed this comment, I hope it is more clear now.

mandrys: OK, thanks, now I see what you meant.

mandrys (approved):

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,49 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
+                      ? new AST::AstNode(AST::AST_SUB, AST::AstNode::mkconst_int(*dimensions_it - 1, true, 32), (*it)->children[0]->clone())
+                      : (*it)->children[0]->clone();
+        // calculate which struct we selected
+        auto move_offset = new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), elem);
+        // move our expanded dot to currently selected struct
+        expanded->children[0] = new AST::AstNode(AST::AST_ADD, move_offset->clone(), expanded->children[0]);
+        expanded->children[1] = new AST::AstNode(AST::AST_ADD, move_offset, expanded->children[1]);
+        // next range, multiply struct_size_int by whole dimension size
+        struct_size_int *= *dimensions_it;
+        // move iterator to next dimension width
```
mglb, line 1004: Obvious comment. You can optionally mention why you move iterator by 2, like "multirange_dimensions stores interleaved offset and size. Move to next dimension's size"

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,49 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
+    // but wiretype ranges contains also struct range that is already expanded in 'expand_dot'
+    // we need to find a way to calculate size of wire ranges without struct range here to enable this assert
+    // const auto wire_node_packed_ranges_size = wire_node->attributes[UhdmAst::packed_ranges()]->children.size();
+    // const auto wire_node_ranges_size = wire_node_packed_ranges_size + wire_node_unpacked_ranges_size;
+    // log_assert(struct_ranges.size() == (wire_node_ranges_size - 1));
+
+    // Get size of single structure
+    int struct_size_int = get_max_offset_struct(struct_node) + 1;
+    // Get reverse iterator of wire dimensions
+    auto dimensions_it = wire_node->multirange_dimensions.rbegin();
```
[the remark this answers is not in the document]
mglb: Also, please rename it to something like `wire_dimension_size_it`,

systemverilog-plugin/UhdmAst.cc
```diff
@@ -962,29 +962,49 @@ static AST::AstNode *convert_dot(AST::AstNode *wire_node, AST::AstNode *node, AS
...
+        // if unpacked range, select from back;
+        auto elem = is_unpacked_range
+                      ? new AST::AstNode(AST::AST_SUB, AST::AstNode::mkconst_int(*dimensions_it - 1, true, 32), (*it)->children[0]->clone())
+                      : (*it)->children[0]->clone();
+        // calculate which struct we selected
+        auto move_offset = new AST::AstNode(AST::AST_MUL, AST::AstNode::mkconst_int(struct_size_int, true, 32), elem);
+        // move our expanded dot to currently selected struct
+        expanded->children[0] = new AST::AstNode(AST::AST_ADD, move_offset->clone(), expanded->children[0]);
+        expanded->children[1] = new AST::AstNode(AST::AST_ADD, move_offset, expanded->children[1]);
+        // next range, multiply struct_size_int by whole dimension size
```
mglb, line 1002: Redundant comment.

mglb (commented):
Logic looks good, but this was quite hard to follow. What (I think) would make it more readable:
- brief description how the input (`expand_dot` output) looks, and what has to be done. Really short, kind of "We have something like this, and have to extract this and that and glue it together into a tree like this".
- remove comments with tautology
- more precise names, which would make few comments redundant.

PR #533, opened 2023-06-19
systemverilog-plugin: fix convert_range function

kamilrakoczy:
This fixes multiranges for wires that doesn't start with 0.

yosys-systemverilog run: https://github.com/antmicro/yosys-systemverilog/actions/runs/5310972213
UHDM-integration-tests: https://github.com/chipsalliance/UHDM-integration-tests/pull/745, https://github.com/chipsalliance/UHDM-integration-tests/pull/744

All failing formal verification tests in yosys-systemverilog are false-negative. I've verified that they produce the same AST:
```
UHDM-integration-tests:MultiAssignmentPatternOfConcat/top.sv
UHDM-integration-tests:MemoryPort/top.sv
```

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -195,6 +195,18 @@ static void delete_children(AST::AstNode *node)
     node->children.clear();
 }
 
+static std::vector<AST::AstNode *> get_ranges(AST::AstNode *id)
+{
+    log_assert(!id->children.empty());
+    std::vector<AST::AstNode *> ranges;
+    for (auto c : id->children) {
+        if (c->type == AST::AST_RANGE) {
+            ranges.push_back(c);
+        }
+    }
+    return ranges;
+}
+
 static void simplify_sv(AST::AstNode *current_node, AST::AstNode *parent_node);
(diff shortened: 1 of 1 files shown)


systemverilog-plugin/UhdmAst.cc
```diff
@@ -168,6 +168,18 @@ static void delete_children(AST::AstNode *node)
     node->children.clear();
 }
 
+static std::vector<AST::AstNode *> get_ranges(AST::AstNode *id)
+{
+    log_assert(!id->children.empty());
+    std::vector<AST::AstNode *> ranges;
+    for (auto c : id->children) {
+        if (c->type == AST::AST_RANGE) {
+            ranges.push_back(c);
```
wsipak, line 177: From what I see this function copies pointers, right?
I think we should either `clone` the nodes, or, If we don't need to access them from the original node, erase the pointers from `id->children` after the pointers are copied to a vector to be returned from the function.

wsipak (changes requested):

mglb:
PR moved to Synlig repository: https://github.com/chipsalliance/synlig/pull/2072

PR #538, opened 2023-08-17
Allows Union Types in Parameter Fields

QuantamHD:
If a union type was referenced as a parameter this would fail.

### DIFF
--- a/systemverilog-plugin/UhdmAst.cc
+++ b/systemverilog-plugin/UhdmAst.cc
@@ -4835,6 +4835,7 @@ void UhdmAst::process_parameter()
             shared.report.mark_handled(typespec_h);
             break;
         }
+        case vpiUnionTypespec:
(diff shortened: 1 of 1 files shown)


QuantamHD:
@kamilrakoczy

mithro (approved):
LGTM

Would be nice to have a test.

mglb (approved):
yosys-systemverilog passes.
https://github.com/antmicro/yosys-systemverilog/actions/runs/5899780880

Can you add test for this to https://github.com/chipsalliance/UHDM-integration-tests? You can base it on some recent test ([example](https://github.com/chipsalliance/UHDM-integration-tests/tree/master/tests/ImportedFunctionCallAsInstantiatedModulePortValueInsideGenFor)) and change just the `.sv` file contents.

QuantamHD:
Done.

PR #541, opened 2023-08-29
Undef a macro that clashes with a UHDM class.

hzeller:
The class `cover` in UHDM will not compile well if the `cover` macro from yosys kernel/log.h is defined. Undef it.


tgorochowik (approved):

tgorochowik:
It would probably be nice to have testcase that shows this error in CI, but let's not wait if it's a blocker for you.

PR #545, opened 2023-09-06
Bump plugins version

kbieganski:
This is the last release with the SV plugin included.


kbieganski:
This is the last release with the SV plugin included.

hzeller (approved):

